PR #36, opened 2014-12-02
Update RemoteConnection.java

larryone:
Path at the end of the JMXServiceURL needs to be configurable
While the path in the vast majority of cases is "jmxrmi", this is not always the case:
http://karaf.apache.org/manual/latest/users-guide/monitoring.html
This change allows integration with Apache Karaf containers, while maintaining "jmxrmi" as a default value.


remh:
Looks good to me!
Thanks!

PR #39, opened 2015-02-03
Support list of filters + tests

### LINKED ISSUE #20: Be more flexible on configuration
- Support wildcards
- Support list of filters instead of simple filters.
- ...

Use case:
I want to exclude a few column families for the cassandra integration.

yannmh:
Fix #20

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Filter.java
+++ b/src/main/java/org/datadog/jmxfetch/Filter.java
@@ -1,0 +1,108 @@
+package org.datadog.jmxfetch;
+
+import java.util.LinkedHashMap;
+import java.util.ArrayList;
+import java.util.Set;
+import java.lang.ClassCastException;
+
+
+class Filter {
+    LinkedHashMap<String, Object> filter;
+
+    /**
+     * A simple class to manipulate include/exclude filter elements more easily
+     * A filter may contain:
+     * - A domain (key: 'domain')
+     * - Bean names (key: 'bean' or 'bean_name')
+     * - Attributes (key: 'attribute')
+     * - Additional bean parameters (other keys)
+     */
+
+    @SuppressWarnings("unchecked")
+    public Filter(Object filter) {
+        LinkedHashMap<String, Object> castFilter;
+        if (filter != null) {
+            castFilter = (LinkedHashMap<String, Object>) filter;
+        } else{
+            castFilter = new LinkedHashMap<String, Object>();
+        }
+        this.filter = castFilter;
+    }
+
+    public String toString() {
+        return this.filter.toString();
+    }
+
+    public Set<String> keySet() {
+        return filter.keySet();
+    }
+
+
+    @SuppressWarnings({ "unchecked", "serial" })
+    private static ArrayList<String> toStringArrayList(final Object toCast) {
+        // Return object as an ArrayList wherever it's defined as
+        // list or not
+        //
+        // ### Example
+        // object:
+        //                  - firstValue
+        //                  - secondValue
+        // ### OR
+        // object: singleValue
+        // ###
+        try{
+            return (ArrayList<String>) toCast;
+        } catch (ClassCastException e){
+            return new ArrayList<String>() {{
+                add(((String) toCast));
+            }};
+        }
+    }
+
+
+    public ArrayList<String> getBeanNames() {
+        if (isEmptyBeanName()){
+            return new ArrayList<String>();
+        }
+        final Object beanNames = (filter.get("bean") != null) ? filter.get("bean") : filter.get("bean_name");
+        // Return bean names as an ArrayList wherever it's defined as
+        // list or not
+        //
+        // ### Example
+        // bean:
+        //                  - org.apache.cassandra.db:type=Caches,keyspace=system,cache=HintsColumnFamilyKeyCache
+        //                  - org.datadog.jmxfetch.test:type=type=SimpleTestJavaApp
+        // ### OR
+        // bean: org.datadog.jmxfetch.test:type=type=SimpleTestJavaApp
+        // ###
+        return toStringArrayList(beanNames);
+    }
+
+    public String getDomain() {
+        return (String) filter.get("domain");
+    }
+
+    public Object getAttribute() {
+        return filter.get("attribute");
+    }
+
+    public ArrayList<String> getParameterValues(String parameterName) {
+        // Return bean attributes values as an ArrayList wherever it's defined as
+        // list or not
+        //
+        // ### Example
+        // bean_parameter:
+        //                  - exampleType1
+        //                  - exampleType2
+        // ### OR
+        // bean_parameter: onlyOneType
+        // ###
+        final Object beanValues = filter.get(parameterName);
+        return toStringArrayList(beanValues);
+    }
+
+    public boolean isEmptyBeanName() {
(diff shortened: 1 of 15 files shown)


src/main/java/org/datadog/jmxfetch/Filter.java
```diff
@@ -0,0 +1,108 @@
...
+     * A simple class to manipulate include/exclude filter elements more easily
+     * A filter may contain:
+     * - A domain (key: 'domain')
+     * - Bean names (key: 'bean' or 'bean_name')
+     * - Attributes (key: 'attribute')
+     * - Additional bean parameters (other keys)
+     */
+
+    @SuppressWarnings("unchecked")
+	public Filter(Object filter) {
```
remh: That's a weird indentation

yannmh:
Rebased on JMXFetch 0.4.1

src/main/java/org/datadog/jmxfetch/Filter.java
```diff
@@ -0,0 +1,108 @@
...
+            }};
+        }
+    }
+
+
+    public ArrayList<String> getBeanNames() {
+        if (isEmptyBeanName()){
+            return new ArrayList<String>();
+        }
+        final Object beanNames = (filter.get("bean") != null)?filter.get("bean"):filter.get("bean_name");
```
remh: nitpick but you should add spaces when using ternary conditions. It would be more readable

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -44,22 +45,44 @@
...
+        for (String param : beanParameters) {
+            String[] paramSplit = param.split(":");
+            beanParams.put(new String(paramSplit[0]), new String(paramSplit[1]));
+        }
+
+        return beanParams;
+    }
+
+
+    private static LinkedList<String> beanParametersToTagsList(String instanceName, String domain, String beanParameters, HashMap<String, String> instanceTags) {
```
remh: This should probably be called:

getDefaultTags(

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -44,22 +45,44 @@
...
+
         // A bean name is formatted like that: org.apache.cassandra.db:type=Caches,keyspace=system,cache=HintsColumnFamilyKeyCache
         // i.e. : domain:bean_parameter1,bean_parameter2
-        String[] split = this.beanName.split(":");
-        this.domain = split[0];
-        this.attributeName = attribute.getName();
+        String[] splitBeanName = this.beanName.split(":");
+        String domain = splitBeanName[0];
+        String beanParameters = splitBeanName[1];
+        LinkedList<String> beanParametersList = beanParametersToTagsList(instanceName, domain, beanParameters, instanceTags);
```
remh: This should probably be called defaultTags

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -44,22 +45,44 @@
...
+        HashMap<String, String> beanParametersHash = beanParameterstoHash(beanParametersList);
```
remh: Nitpick but you should use verbs in function names.

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -100,39 +123,35 @@ Object getJmxValue() throws AttributeNotFoundException, InstanceNotFoundExceptio
     }
 
     boolean matchDomain(Configuration conf) {
-        LinkedHashMap<String, Object> include = conf.getInclude();
```
remh: another nitpick but something like:
```java
String confDomain = conf.getInclude().getDomain();
return confDomain == null || confDomain.equals(domain);
```
would be a bit better

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -100,39 +123,35 @@ Object getJmxValue() throws AttributeNotFoundException, InstanceNotFoundExceptio
...
-        return include.get("domain") == null || include.get("domain").equals(domain);
+        Filter include = conf.getInclude();
+        return include.getDomain() == null || include.getDomain().equals(domain);
     }
 
     boolean excludeMatchDomain(Configuration conf) {
-        LinkedHashMap<String, Object> exclude = conf.getExclude();
```
remh: Same thing here

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -100,39 +123,35 @@ Object getJmxValue() throws AttributeNotFoundException, InstanceNotFoundExceptio
...
-                String[] paramSplit = param.split(":");
-                beanParams.put(new String(paramSplit[0]), new String(paramSplit[1]));
-            }
-
-            if (exclude.get(bean_attr).equals(beanParams.get(bean_attr))) {
-                return true;
-            }
-        }
+       Filter exclude = conf.getExclude();
+       ArrayList<String> beanNames = exclude.getBeanNames();
```
remh: Same here

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -100,39 +123,35 @@ Object getJmxValue() throws AttributeNotFoundException, InstanceNotFoundExceptio
...
+       }
+
+       for (String bean_attr : exclude.keySet()) {
+           if (EXCLUDED_BEAN_PARAMS.contains(bean_attr)) {
+               continue;
+           }
+
+           ArrayList<String> beanValues = exclude.getParameterValues(bean_attr);
+           for (String beanVal : beanValues) {
+               if (beanParameters.get(bean_attr) != null && beanParameters.get(bean_attr).equals(beanVal)) {
```
remh: You should get the first condition out of the loop, and then just test the presence of the attribute in the list.

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -199,13 +210,18 @@ boolean matchBean(Configuration configuration) {
...
-            HashMap<String, String> beanParams = new HashMap<String, String>();
-            for (String param : this.defaultTags) {
-                String[] paramSplit = param.split(":");
-                beanParams.put(new String(paramSplit[0]), new String(paramSplit[1]));
-            }
+            ArrayList<String> beanValues = include.getParameterValues(bean_attr);
 
-            if (beanParams.get(bean_attr) == null || !beanParams.get(bean_attr).equals(include.get(bean_attr))) {
+            boolean matchBeanAttr = false;
+            for (String beanVal : beanValues) {
```
remh: Same here

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -199,13 +210,18 @@ boolean matchBean(Configuration configuration) {
...
+                if (beanParameters.get(bean_attr) == null || !beanParameters.get(bean_attr).equals(beanVal)) {
+                    continue;
+                }
+                matchBeanAttr = true;
```
remh: You could return true right there.

PR #50, opened 2015-04-03
Seth/cassandra metrics update

### LINKED ISSUE #6: Cassandra metrics change starting with version 1.2
Cassandra now exposes both:

```
org.apache.cassandra.db.ColumnFamilies <<<<<<<<<<< old
org.apache.cassandra.metrics.ColumnFamily <<<<<<<<<<< new
```

See [this](https://wiki.apache.org/cassandra/Metrics) for more details on the new style.

One key change is that the old format was:
1. CF is an MBean
2. CF has a bunch of attributes that map to metrics
while the new format has
1. CF is a scope
2. the MBean is the metric name
3. the attribute is named "Value"

sethrosenblum:
This allows jmxfetch to properly map the new metrics structure from Cassandra to something more usable.  This is intended to fix #6.

[Cassandra Metrics Reference](http://wiki.apache.org/cassandra/Metrics)

This change does several things:
* Maintains parity for the deprecated Cassandra namespaces
* Remaps metric names to `cassandra.$metric` or `cassandra.$metric.$attribute`
* Drops `name:$name` tag because of redundancy
* Renames `scope:$scope` to `$type:$scope` so users can continue to use 'columnfamily:mytable`, for instance

I also did some refactoring around moving post processing back into the processing of a metric and reusing domain.  I can break that into a separate Pull Request if that's preferable.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/JMXSimpleAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXSimpleAttribute.java
@@ -102,7 +102,7 @@ public class JMXSimpleAttribute extends JMXAttribute {
     }
 
     private String getCassandraAlias() {
-        if (getDomain().equals("org.apache.cassandra.metrics")) {
+        if (getDomain().equals(CASSANDRA_DOMAIN)) {
             Map<String, String> beanParameters = getBeanParameters();
             String type = beanParameters.get("type");
             String metricName = beanParameters.get("name");
--- a/src/main/java/org/datadog/jmxfetch/JMXSimpleAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXSimpleAttribute.java
@@ -91,11 +91,17 @@ public class JMXSimpleAttribute extends JMXAttribute {
             alias = attribute.get(getAttribute().getName()).get("alias");
         } else if (conf.get("metric_prefix") != null) {
             alias = conf.get("metric_prefix") + "." + getDomain() + "." + getAttributeName();
+        } else if (getDomain().startsWith("org.apache.cassandra")) {
+            alias = getCassandraAlias();
         } else {
             alias = "jmx." + getDomain() + "." + getAttributeName();
         }
         alias = convertMetricName(alias);
         return alias;
+    }
+
+    private String getCassandraAlias() {
+        return getDomain().replace("org.apache.", "") + "." + getAttributeName();
     }
 
     private String getMetricType() {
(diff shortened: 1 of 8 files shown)


yannmh:
Great ! Thanks @sethrosenblum for working on this.
I'll review it. Let's aim for the 5.4.0 agent release.

src/main/java/org/datadog/jmxfetch/JMXSimpleAttribute.java
```diff
@@ -90,14 +91,31 @@ private String getAlias() {
...
+            alias = getCassandraAlias();
         } else {
-            alias = "jmx." + getBeanName().split(":")[0] + "." + getAttributeName();
+            alias = "jmx." + getDomain() + "." + getAttributeName();
         }
         alias = convertMetricName(alias);
         return alias;
     }
 
+    private String getCassandraAlias() {
```
yannmh: Is there any reason why we don't want the same approach for `JMXComplexAttribute` ?

yannmh:
Do you think you can add some tests ?

For instance we could check that we have the expected metric name, tags from a given bean (and ensure backward compatibility in the meantime).

src/main/java/org/datadog/jmxfetch/JMXSimpleAttribute.java
sethrosenblum: Cassandra doesn't produce any complex attributes.

sethrosenblum:
I had been hesitent to add a mock cassandra bean, as our other tests had all used a fake java bean.  I'll see if I can make something to exercise the code path.

yannmh:
Hey @sethrosenblum ,

As we are changing Cassandra metric names, we are likely to introduce some backward imcompatibilities/discontinuities for some users. Are we ok with that ?

sethrosenblum:
If a user continues to use their existing cassandra.yaml with the new jmxfetch, their metric names should stay the same, and they shouldn't see any difference.  When we release a new cassandra.yaml to take advantage of the new cassandra namespace, then users will see a name change as the metric data being presented is quite different.

yannmh:
Custom metrics defined in `cassandra.yaml` without an `alias` would still be impacted. 
I remember having seen some 'cassandra.metrics.$value' style metric name among users  (cc @amankapur ).

yannmh:
It's definitely breaking backward compatibility when users have these 'new-style' metrics set as custom but no aliases (I counted at least 9 users).

We can bump JMXFetch version to `1.0.0`,  and also reach the impacted users as it would be embedded by default with Datadog agent 5.4.0.
@remh any thoughts ?

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -60,8 +60,6 @@ public void sendMetrics(LinkedList<HashMap<String, Object>> metrics, String inst
             // We need to edit metrics for legacy reasons (rename metrics, etc)
             HashMap<String, Object> metric = new HashMap<String, Object>(m);
 
-            postProcess(metric);
```
yannmh: Should this method be removed ?

yannmh:
Let's postpone theses changes to the 5.5.0 agent release: it'll give us more time to handle the transition.

Also, it may worth a shot double checking the (non?)existence of `JMXComplexAttribute`s for Cassandra metrics.

yannmh:
JMX attribute types over my Cassandra instance: 
```
'double',
'java.lang.Object',
'java.lang.String',
'long'
```

None of them is part of `JMXComplexAttributes `definition.

yannmh:
Closing in favor of https://github.com/DataDog/jmxfetch/pull/79

PR #51, opened 2015-04-08
Only send instance name in service check tags

remh:
@elafarge @conorbranagan @yannmh this will ensure that we can group the service checks properly in the backend.


conorbranagan:
Looks good to me, thanks! Are there custom tags we should be adding too?

remh:
Not sure we should do that in this version though. I'd like to see a bit more how it behaves first.

yannmh:
Looks good to me. Thanks @remh

PR #53, opened 2015-04-16
[activemq58] Proper prefix on service check

elafarge:
I hacked the StatsDReporter to send the status of activemq.can_connect
instead of activemq_58.can_connect. Hope it was the right place to do
so.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
@@ -50,8 +50,9 @@ public class StatsdReporter extends Reporter {
             init();
         }
 
-        ServiceCheck sc = new ServiceCheck(String.format("%s.can_connect", checkName),
-            this.statusToInt(status), message, hostname, tags);
+        String dataName = Reporter.formatDataName(String.format("%s.can_connect", checkName));
+        ServiceCheck sc = new ServiceCheck(dataName, this.statusToInt(status), message, hostname, tags);
+
         statsDClient.serviceCheck(sc);
     }
 
--- a/pom.xml
+++ b/pom.xml
@@ -11,6 +11,7 @@
 
 	<properties>
 		<commons-io.version>2.4</commons-io.version>
+		<commons-lang.version>2.6</commons-lang.version>
 		<guava.version>17.0</guava.version>
 		<java-dogstatsd-client.version>2.1.0</java-dogstatsd-client.version>
 		<jcommander.version>1.35</jcommander.version>
@@ -51,6 +52,11 @@
 			<groupId>commons-io</groupId>
 			<artifactId>commons-io</artifactId>
 			<version>${commons-io.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>commons-lang</groupId>
+			<artifactId>commons-lang</artifactId>
+			<version>${commons-lang.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>com.datadoghq</groupId>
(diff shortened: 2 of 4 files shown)


elafarge:
DO NOT merge : I probably found a way better solution

src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -28,7 +28,7 @@ protected void sendMetricPoint(String metricName, double value, String[] tags) {
             this.statsDClient.stop();
             init();
         }
-        statsDClient.gauge(metricName, value, tags);
+        statsDClient.gauge(StatsdReporter.formatDataName(metricName), value, tags);
```
yannmh: `metricName` should be alias names defined in user configurations. I don't think we want/need to format it.

elafarge: Yeah, I know that. The problem is that the service check name isn't defined in user configurations :/

My point here was just to enable us to stop getting the service check prefix from the configuration file and therefore being able to have versioned configuration file without getting versioned prefixes in our backend (activemq.yaml, activemq_58.yaml, activemq_62.yaml...). In the first place I just did a dirty `if(checkName == 'activemq_58') checkName = 'activemq';` do you think this would be more suitable ? Or maybe we could format the metricName when it's read from the yaml file name instead of doing it every time the metrics'service checks are sent.

elafarge: Or actually make it so the service check name is also read from the configuration, just like the metrics are... what do you think ?

src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -50,11 +50,29 @@ public void sendServiceCheck(String checkName, String status, String message,
...
-        ServiceCheck sc = new ServiceCheck(String.format("%s.can_connect", checkName),
-            this.statusToInt(status), message, hostname, tags);
+        //ActiveMQ > 5.8 hack
+        String dataName = StatsdReporter.formatDataName(String.format("%s.can_connect", checkName));
+        ServiceCheck sc = new ServiceCheck(dataName, this.statusToInt(status), message, hostname, tags);
+
         statsDClient.serviceCheck(sc);
     }
 
+    public static String formatDataName(String name){
```
yannmh: This should probably be moved to `Reporter.java` so it can be used with `ConsoleReporter` too (used by default with agent `jmx` commands).

I would also move formating logic to a new method in `Reporter.java`: it would trigger `sendServiceCheck` (private method ?) directly with the right, formatted parameters so we avoid duplicating the code among reporters.
Let me know what you think :smiley:

src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -50,11 +50,29 @@ public void sendServiceCheck(String checkName, String status, String message,
...
+        String[] chunks = name.split("\\.");
+        
+        // Escaping the prefix name 
+        chunks[0] = chunks[0].replaceAll("[A-Z0-9:_\\-]", "");
+
+        // Putting the chunks together (there's nothing in the standard library to do that :o )
```
yannmh: Did you have a change to try [StringUtils.join](http://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#join%28java.util.Collection,%20char%29) from Apache Commons Lang ? That would be way easier

yannmh:
Great work, thanks @elafarge.

I wrote a few suggestions. Please let me know what you think about it.

elafarge:
Thanks :) I just finished implementing your suggestions. It looks better now :)

elafarge:
Moved there https://github.com/DataDog/jmxfetch/pull/56, closing the PR.

PR #54, opened 2015-04-17
[service-checks] JMX Serv. checks sent to dd-agent

elafarge:
The agents gets, through a jmx_status.yaml file, the service checks
sent for each checks. This value is the one displayed when running
datadog-agent info.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -284,11 +284,13 @@ public class App {
     private void reportStatus(AppConfig appConfig, Reporter reporter, Instance instance,
                               int metricCount, String message, String status) {
         String checkName = instance.getCheckName();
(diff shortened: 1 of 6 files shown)


yannmh:
Thanks for taking care of it @elafarge !

Quite nitpicky: can we count service checks from `App` class instead of `Reporter` like we are already doing for metrics ? It would avoid breaking the 'code symmetry, and makes the code lighter. 
For instance, we could add a `serviceChecksCount` parameter to `addInstanceStats ` (which seems to be always equals to `1`).
https://github.com/DataDog/jmxfetch/blob/etienne/service-check-count-fix/src/main/java/org/datadog/jmxfetch/App.java#L293

elafarge:
Should be ready to merge. This being said, this branch is included in https://github.com/DataDog/jmxfetch/pull/56. Merging the latter would therefore put the two commits from this branch into master without having to merge it :)

yannmh:
Merging :dancers:

PR #58, opened 2015-05-26
File to trigger JMXFetch exit

yannmh:
Trigger JMXFetch exit when a specified file is created.
Usage: `--exit_file_location` or `-e` with the absolute path of the trigger
file to watch.


remh:
Looks good to me

PR #59, opened 2015-05-26
Rename 'host' bean parameter to 'bean_host' in tags

olivielpeau:
@yannmh Fixes the case when a bean parameter named 'host' overrides the actual 'host' tag.

### DIFF
--- a/src/test/java/org/datadog/jmxfetch/TestApp.java
+++ b/src/test/java/org/datadog/jmxfetch/TestApp.java
@@ -48,7 +48,7 @@ public class TestApp {
     public void testBeanTags() throws Exception {
         // We expose a few metrics through JMX
         MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
-        ObjectName objectName = new ObjectName("org.datadog.jmxfetch.test:type=SimpleTestJavaApp,scope=CoolScope");
+        ObjectName objectName = new ObjectName("org.datadog.jmxfetch.test:type=SimpleTestJavaApp,scope=CoolScope,host=localhost");
         SimpleTestJavaApp testApp = new SimpleTestJavaApp();
         mbs.registerMBean(testApp, objectName);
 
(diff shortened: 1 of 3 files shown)


yannmh:
Looks great, thanks @olivielpeau !

Shall we merge it for the 5.4 agent release ? cc @remh

olivielpeau:
@remh This PR shoud be mergeable, should I open a PR on https://github.com/DataDog/documentation  to document the change ?

remh:
It would be simpler and more efficient to just check the value of the key here and replace it:
https://github.com/DataDog/jmxfetch/blob/olivielpeau/rename-host-parameter/src/main/java/org/datadog/jmxfetch/JMXAttribute.java#L84

olivielpeau:
I've thought about it but one problem I had with it is that it would change not only the name of the tag that is sent, but also the name of the bean parameter itself in `JMXAttribute`, and that could lead to confusing things when we try to match the `JMXAttribute` with the configuration.

That's why I did it in two steps: first deal with the bean parameters (and keep the `host` bean parameter unchanged) and then deal with the default tags (and change the name of the tag to `bean_host`).

Actually I think there are still a few confusing things about the way we don't clearly separate bean parameters from default tags. For instance I think the `instanceTags` shouldn't be added to the bean parameters list, but only to the default tags.

src/test/java/org/datadog/jmxfetch/TestApp.java
```diff
@@ -74,11 +74,13 @@ public void testBeanTags() throws Exception {
         	Set<String> tagsSet = new HashSet<String>(Arrays.asList(tags));
 
         	// We should find bean parameters as tags
-        	assertEquals(4, tags.length);
+            assertEquals(5, tags.length);
```
remh: Can you fix the indentation here please ?

src/test/java/org/datadog/jmxfetch/TestApp.java
```diff
@@ -74,11 +74,13 @@ public void testBeanTags() throws Exception {
...
         	assertEquals(true, tagsSet.contains("type:SimpleTestJavaApp"));
         	assertEquals(true, tagsSet.contains("scope:CoolScope"));
         	assertEquals(true, tagsSet.contains("instance:jmx_test_instance"));
         	assertEquals(true, tagsSet.contains("jmx_domain:org.datadog.jmxfetch.test"));
+            // Special case of the 'host' parameter which tag is renamed to 'bean_host'
+            assertEquals(true, tagsSet.contains("bean_host:localhost"));
```
remh: Same here

remh:
Good point.

Looks good to me besides the indentation issues. Can you fix them and then merge the PR ? Thanks! 

:boom:

olivielpeau:
Fixed the indentation issues, merging!

PR #62, opened 2015-06-03
Support bean names that have an attribute with an empty value

olivielpeau:
cc @remh 
The java docs allow MBean attributes to have empty values (see http://docs.oracle.com/javase/6/docs/api/javax/management/ObjectName.html).
Before this fix JMXAttribute was assuming that values were not empty and would throw an unexpected exception upon parsing such a bean name (https://datadog.zendesk.com/agent/tickets/27140).

This fixes the issue. Attributes with empty values are filtered out of the default tags.


olivielpeau:
@remh 
Changed the behavior to include empty-value attributes in the default tags.

yannmh:
Added a note to 0.70 CHANGELOG. Please let me know when it's merged

olivielpeau:
@remh 
Removed the colon at the end of empty-value tags.

PR #63, opened 2015-07-22
Limit MBeans query to certain scopes

yannmh:
JMXFetch is currently fetching all available beans regardless of the
user configuration. This might cause a high memory load when the
application is exposing a lot of metrics (thus OOM issues).

This change is the first step towards a bigger JMXFetch memory
consumption improvement: the idea is to find, among the user defined
beans, some common patterns to limit the scopes to query.

For instance,
```
- org.datadog.jmxfetch.test:scope=sameScope,param=sameParam,type=sameType
- org.datadog.jmxfetch.test:scope=sameScope,param=notTheSameParam,type=sameType
```
would induce a query on `org.datadog.jmxfetch.test:scope=sameScope,type=sameType,*`
instead of `*:*`.

At the moment, only one scope per domain name is computed: it's kind of
the greatest common divisor/pattern of the beans.

Ideally, this should be improved to recognize multiple scopes among the
same domain name.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Configuration.java
+++ b/src/main/java/org/datadog/jmxfetch/Configuration.java
@@ -1,6 +1,14 @@
 package org.datadog.jmxfetch;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.Map.Entry;
+import java.util.Set;
 
 public class Configuration {
 
@@ -9,7 +17,9 @@ public class Configuration {
     private Filter exclude;
 
     /**
-     * A simple class to access configuration elements more easily
+     * Access configuration elements more easily
+     *
+     * Also provides helper methods to extract common information among filters.
      */
     public Configuration(LinkedHashMap<String, Object> conf) {
         this.conf = conf;
@@ -20,7 +30,6 @@ public class Configuration {
     public LinkedHashMap<String, Object> getConf() {
         return conf;
     }
-
     public Filter getInclude() {
         return include;
     }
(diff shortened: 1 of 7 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -263,9 +263,26 @@ private void getMatchingAttributes() {
...
+        this.beans = new HashSet<ObjectInstance>();
+        try {
+            LinkedList<String> beanScopes = getBeansScopes();
+            for (String scope : beanScopes) {
+                ObjectName name = new ObjectName(scope);
+                this.beans.addAll(connection.queryMBeans(name));
+            }
+        }
+        catch (Exception e) {
+            e.printStackTrace();
```
olivielpeau: Bit of a nitpick, but could we also log an error so that the exception is not completely silent when JMXFetch's stderr is not redirected to a log?
Something like `LOGGER.error("Unable to compute common bean scope, querying all beans as a fallback")`

remh: (not a nitpick and it would be better to catch a more specific exception)

yannmh: Good idea. 

Otherwise I can't think of a specific exception that i want to catch. I added this very generic `try {} catch(){}` statement in order to fallback in the previous behavior in case anything was turning wrong. I should maybe remove it completely.

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -355,4 +355,8 @@ String getBeanName() {
     String getAttributeName() {
         return attributeName;
     }
+
+    public static List<String> getExcludeBeanParams(){
```
olivielpeau: Just to be consistent, could we rename this method `getExcludedBeanParams`? :)  (ie `Exclude` --> `Excluded`)

src/main/java/org/datadog/jmxfetch/Configuration.java
```diff
@@ -33,4 +42,212 @@ public String toString() {
...
+    public static LinkedList<String> getGreatestCommonScopes(LinkedList<Configuration> configurationList){
+        LinkedList<Configuration> includeConfigList = getIncludeConfigurationList(configurationList);
+        HashMap<String, LinkedList<Filter>> includeFiltersByDomain = getIncludeFiltersByDomain(includeConfigList);
+        HashMap<String, Set<String>> parametersIntersectionByDomain = getCommonBeanKeysByDomain(includeFiltersByDomain);
+        HashMap<String, LinkedHashMap<String, String>> commonBeanScopeByDomain = getCommonScopeByDomain(parametersIntersectionByDomain, includeFiltersByDomain);
+
+        LinkedList<String> result = new LinkedList<String>();
+        Iterator<Entry<String, LinkedHashMap<String, String>>> commonBeanScopeItr = commonBeanScopeByDomain.entrySet().iterator();
+        while(commonBeanScopeItr.hasNext()){
+            Entry<String, LinkedHashMap<String, String>> beanScopeEntry = commonBeanScopeItr.next();
```
olivielpeau: You don't modify the `HashMap` that is being iterated on here so you use a `for` loop :wink: 

```
for (Entry<String, LinkedHashMap<String, String>> beanScopeEntry: commonBeanScopeByDomain.entrySet()) {
```

Also applies to lines 129, 164 and 215.

src/main/java/org/datadog/jmxfetch/Configuration.java
```diff
@@ -33,4 +42,212 @@ public String toString() {
...
+            String domain = (String) beanScopeEntry.getKey();
+            LinkedHashMap<String, String> beanScope = (LinkedHashMap<String, String>) beanScopeEntry.getValue();
```
olivielpeau: There's no need to cast the values returned by `getKey()` and `getValue()` since they're already of type `String` and `LinkedHashMap<String, String>` respectively.

Also applies to lines 131-132, 166-167, 217-218.

src/main/java/org/datadog/jmxfetch/Configuration.java
```diff
@@ -33,4 +42,212 @@ public String toString() {
...
+                    ArrayList<String> parameterValues = f.getParameterValues(parameter);
+
+                    if (parameterValues.size() != 1 || (commonValue != null && !commonValue.equals(parameterValues.get(0)))) {
+                        hasCommonValue = false;
+                        continue;
+                    }
+
+                    commonValue = parameterValues.get(0);
+
+                }
```
olivielpeau: Might just be a personal taste, but I'd also prefer a `for` loop here:
```
for (Filter f : filters) {
    ArrayList<String> parameterValues = f.getParameterValues(parameter);

    if (parameterValues.size() != 1 || (commonValue != null && !commonValue.equals(parameterValues.get(0)))) {
        hasCommonValue = false;
        break;
    }

    commonValue = parameterValues.get(0);
}
```

olivielpeau:
Looks great! :+1: 

Nothing really important in my comments except for the exception thing.

yannmh:
Thanks for the feedback. 
I adressed your comments. It should be ready to merge now :rocket:. 

This closes the *JMXFetch v0.8.0* release: I'll write the *CHANGELOG* and submit the new artifact to *dd-agent*.

PR #65, opened 2015-07-29
Remove host name from service checks payloads

olivielpeau:
The host name that is currently sent is the `host` defined in the yaml config for each instance, which is preventing the backend from correctly assigning host tags to the service checks.

This removes it completely ; the forwarder will take care of adding the correct hostname.

BC break: for users who have set up monitors on service checks filtered by `host`, the service check will disappear (because the value of the tag changes with this PR). Currently this value is `localhost` for most users and therefore quite useless, but it could still be used by a few orgs. We should probably reach out to all the orgs that use JMXFetch before we release this.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -290,8 +290,7 @@ public class App {
     private void reportStatus(AppConfig appConfig, Reporter reporter, Instance instance,
                               int metricCount, String message, String status) {
         String checkName = instance.getCheckName();
-        reporter.sendServiceCheck(checkName, status, message, instance.getHostname(),
-                                  instance.getServiceCheckTags());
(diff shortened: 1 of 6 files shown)


conorbranagan:
We could alternatively not pick the `host_name` value if it's a "local" hostname while processing in Sobotka, similar to what we're doing now for `host:val` tags here: https://github.com/DataDog/dogweb/blob/prod/prozess/sobotka/transform.py#L495-L500. This wouldn't require an Agent upgrade either.

olivielpeau:
Thanks @conorbranagan, I didn't know about that!
Actually we could modify only JMXFetch to make it send service checks with the host in a `host` tag (instead of in the `h` metadata field of dogstatsd). The forwarder would then add a `host_name` tag in the service check, and sobotka would either pick the value of `host` (if it's not a "local" one) or of `host_name`.

PR #66, opened 2015-07-31
Replace host metadata in dogstasd payload with host tag

olivielpeau:
Supersedes https://github.com/DataDog/jmxfetch/pull/65.

The host name that's sent is the host that's defined in the yaml config for each instance, which is preventing the backend from correctly assigning host tags to the service checks.

We fix that by sending a `jmx_server` tag in the dogstatsd payload instead of a host_name metadata field. The agent's dogstatsd will then add the agent's hostname.

Also added an `is_jmx` tag to JMXFetch's service checks so that the backend can know which service checks come from JMXFetch.

cc @talwai @conorbranagan

### DIFF
--- a/src/test/java/org/datadog/jmxfetch/TestApp.java
+++ b/src/test/java/org/datadog/jmxfetch/TestApp.java
@@ -481,9 +481,6 @@ public class TestApp {
 
     @Test
     public void testServiceCheckCounter() throws Exception {
-        MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
-        SimpleTestJavaApp testApp = new SimpleTestJavaApp();
-
         AppConfig appConfig = new AppConfig();
         App app = initApp("jmx.yaml", appConfig);
         Reporter repo = appConfig.getReporter();
@@ -494,7 +491,7 @@ public class TestApp {
         // Let's put a service check in the pipeline (we cannot call doIteration()
         // here unfortunately because it would call reportStatus which will flush
         // the count to the jmx_status.yaml file and reset the counter.
-        repo.sendServiceCheck("jmx", Status.STATUS_OK, "This is a test", "jmx_test_instance", null);
+        repo.sendServiceCheck("jmx", Status.STATUS_OK, "This is a test", null);
 
         // Let's check that the counter has been updated
         assertEquals(1, repo.getServiceCheckCount("jmx"));
--- a/src/test/java/org/datadog/jmxfetch/TestApp.java
+++ b/src/test/java/org/datadog/jmxfetch/TestApp.java
@@ -1,11 +1,6 @@
 package org.datadog.jmxfetch;
 
-import com.beust.jcommander.JCommander;
-
-import org.apache.log4j.Level;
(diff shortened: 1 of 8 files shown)


talwai:
cc @remh - what do you think of this approach?

src/test/java/org/datadog/jmxfetch/TestApp.java
```diff
@@ -494,7 +491,7 @@ public void testServiceCheckCounter() throws Exception {
         // Let's put a service check in the pipeline (we cannot call doIteration()
         // here unfortunately because it would call reportStatus which will flush
         // the count to the jmx_status.yaml file and reset the counter.
-        repo.sendServiceCheck("jmx", Status.STATUS_OK, "This is a test", "jmx_test_instance", null);
+        repo.sendServiceCheck("jmx", Status.STATUS_OK, "This is a test", null);
 
         // Let's check that the counter has been updated
         assertEquals(1, repo.getServiceCheckCount("jmx"));
```
yannmh: Can we test the "jmx_server: hostname" tag ? In the meantime, maybe we can start moving those service checks tests in a different file to improve the readibility.

yannmh:
Looks good to me :cake:. 

We should discuss the wording `jmx_server` and the role of the `is_jmx` tag before merging it.

src/test/java/org/datadog/jmxfetch/TestApp.java
olivielpeau: Hmm actually all of our tests rely on the Attach API and thus don't use a `host` in the configuration, so we can't cleanly test the `jmx_server` tag with our current test setup...
We could change our test setup to use JMX Remote instead but that would need some work.

olivielpeau:
Thansk @yannmh for the review!

I've removed the `is_jmx` tag, and moved the service check tests to a separate class. See my comment above about the test on the `jmx_server` tag.
Might be a good idea to change the test setup if we want to test the more conventional JMX Remote instead of the Attach API, don't if/how we could implement that though.

src/test/java/org/datadog/jmxfetch/TestApp.java
yannmh: Oh that's right. 
Switching to JMX Remote is not worth. A better solution would be to be able to unit test the specific `getServiceCheckTags` method, and not instantiate the whole app like we are currently doing. This is what we do for instance with [TestConfiguration](https://github.com/DataDog/jmxfetch/blob/master/src/test/java/org/datadog/jmxfetch/TestConfiguration.java).
Let's see if it can be done easily and decide.

talwai:
LGTM @olivielpeau . Can you merge and build a new jar for the `dd-agent` repo?

olivielpeau:
I've tested the new jar on staging and it works fine: host tags are available during the setup of new monitors on JMXFetch service checks.
Waiting on feedback from the customer to whom we sent this version of the jar, I'll merge the PR tomorrow at the latest.

olivielpeau:
No feedback from the customer unfortunately but merging anyway.

yannmh:
:+1:

PR #67, opened 2015-08-20
Log stack traces instead of only printing them

olivielpeau:
Allows to have the stack traces in the logs on top of being output.


yannmh:
Thanks @olivielpeau. 

Let's merge it once tests pass :watch:

PR #70, opened 2015-08-25
Scope MBeans queries on certain actions only

yannmh:
Do not scope MBeans queries on `list_everything` or
`list_not_matching_attributes`.
Fix regression introduced with #63.


olivielpeau:
Looks good! :)

PR #73, opened 2015-08-28
[sc] Don't send sc. WARNING if too many metrics

elafarge:
We were sending a service check warning for integrations with too many
metrics which is not the behaviour we'd like to have. Even though there
are too many metrics, the `can_connect` service check should be green.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -173,36 +173,44 @@ public class App {
             Instance instance = it.next();
             LinkedList<HashMap<String, Object>> metrics;
             String instanceStatus = Status.STATUS_OK;
+            String scStatus = Status.STATUS_OK;
             String instanceMessage = null;
+            int numberOfMetrics = 0;
+
             try {
                 metrics = instance.getMetrics();
+                numberOfMetrics = metrics.size();
+
+                if (numberOfMetrics == 0) {
+                    instanceMessage = "Instance " + instance + " didn't return any metrics";
+                    LOGGER.warn(instanceMessage);
+                    instanceStatus = Status.STATUS_ERROR;
+                    scStatus = Status.STATUS_ERROR;
+                    brokenInstances.add(instance);
+                } else if (instance.isLimitReached()) {
+                    instanceMessage = "Number of returned metrics is too high for instance: "
+                            + instance.getName()
+                            + ". Please read http://docs.datadoghq.com/integrations/java/ or get in touch with Datadog "
+                            + "Support for more details. Truncating to " + instance.getMaxNumberOfMetrics() + " metrics.";
+
+                    instanceStatus = Status.STATUS_WARNING;
+                    // We don't want to log the warning at every iteration so we use this custom logger.
+                    CustomLogger.laconic(LOGGER, Level.WARN, instanceMessage, 0);
+                }
+
+                if(numberOfMetrics > 0)
+                    reporter.sendMetrics(metrics, instance.getName());
+
             } catch (IOException e) {
-                String warning = "Unable to refresh bean list for instance " + instance;
-                LOGGER.warn(warning, e);
-                reportStatus(appConfig, reporter, instance, 0, warning, Status.STATUS_ERROR);
+                instanceMessage = "Unable to refresh bean list for instance " + instance;
+                LOGGER.warn(instanceMessage, e);
+                instanceStatus = Status.STATUS_ERROR;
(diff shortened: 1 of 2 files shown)


olivielpeau:
Tell me if I'm missing something, but I think all we need to do is to remove this line: https://github.com/DataDog/jmxfetch/blob/master/src/main/java/org/datadog/jmxfetch/App.java#L199

What do you think? :smiley:

elafarge:
Indeed, that would have been faster but...

1. I don't like to have more lines removed than lines added in my Github statistics
2. Actually if we do that, we won't get the warning on the `datadog-agent info` page OR we still wan't to have that one. This they are bound to the same variable at this point in the code I'm afraid it's not possible to just remove that line whilst keeping the WARNING on the `info` page. Correct me if I'm wrong :)

Oh and thanks a lot for your review btw. :)

remh:
@elafarge 
```
"When debugging, novices insert corrective code; experts remove defective code."
Richard Pattis
``` 
:) sometimes removing code is better.

But yeah i agree, we still want the warning to be displayed in the info page (and sent as other agent check warnings in the infra overview if possible).

elafarge:
@yannmh @remh I decoupled the service checks and status. I didn't add a second service check for the alert on the number of metrics but that's something I could do if it appears to be relevant :)

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -201,8 +204,8 @@ public void doIteration() {
                 CustomLogger.laconic(LOGGER, Level.WARN, instanceMessage, 0);
             }
             reporter.sendMetrics(metrics, instance.getName());
-
-            reportStatus(appConfig, reporter, instance, metrics.size(), instanceMessage, instanceStatus);
+            this.reportStatus(appConfig, reporter, instance, metrics.size(), instanceMessage, instanceStatus);
+            this.sendServiceCheck(reporter, instance, instanceMessage, scStatus);
         }
```
yannmh: Can we simplify (factorize) this block with something like :
``` java
while (it.hasNext()) {
	String instanceStatus = Status.STATUS_OK;
	String serviceCheckStatus = Status.STATUS_OK;
	String message = null;
	
	try {
		// Instantiate instanceStatus, serviceCheckStatus, message
		doSomething();
	} catch (Exception) {
		// Modify instanceStatus, serviceCheckStatus, message
		handleException();
	}

	// Factorize instance status and service check report
	this.reportStatus(appConfig, reporter, instance, metrics.size(), instanceMessage, instanceStatus);
	this.sendServiceCheck(reporter, instance, instanceMessage, scStatus);
}
```

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -233,19 +236,23 @@ public void doIteration() {
...
+                this.reportStatus(appConfig, reporter, instance, 0, warning, Status.STATUS_ERROR);
+                this.sendServiceCheck(reporter, instance, warning, Status.STATUS_ERROR);
             } catch (Exception e) {
                 String warning = "Cannot connect to instance " + instance + " for an unknown reason." + e.getMessage();
                 LOGGER.fatal(warning, e);
-                reportStatus(appConfig, reporter, instance, 0, warning, Status.STATUS_ERROR);
+                this.reportStatus(appConfig, reporter, instance, 0, warning, Status.STATUS_ERROR);
+                this.sendServiceCheck(reporter, instance, warning, Status.STATUS_ERROR);
             }
         }
```
yannmh: All of these exceptions are pretty much handled the same way, only the `warning` message differs. I am wondering if we can factorize here:
``` java

try {
	doSomething();
} catch (Exception e) {
	String warning = null;

	if (e instanceof SomeException) {
		warning = "a scary warning"

	} else if( e instanceof AnOtherException){
		warning = "some warning"

	} else {
		// You got the idea
		(...)
	}


	this.reportStatus(appConfig, reporter, instance, 0, warning, Status.STATUS_ERROR);
	this.sendServiceCheck(reporter, instance, warning, Status.STATUS_ERROR);
}
```

yannmh:
Thanks for tackling the issue @elafarge. 

While I think decoupling _service checks_ and _instance status_ is the right thing to do, I am a bit worried about the resulting code duplication. I added some suggestion to refactorize some code paths :art:, please let me know what you think

elafarge:
Thank you so much for you enlightening suggestions @yannmh . I took them into account and updated the PR. Let me know what you think about it ;-)

yannmh:
Thanks @elafarge !

PR #78, opened 2015-10-20
Assign generic alias if we can't find one - was causing NullPointerEx…

truthbk:
…ception's otherwise..


yannmh:
Thanks @truthbk, it looks great.

A single nitpick :art: 
Can we take advantage of thism to refactor the method ? I see some redundancies in the `alias` definition, i.e.
`alias = metric_prefix + "." + getBeanStringName().split(":")[0] + "." + getAttributeName();` appearing multiple times.

truthbk:
Totally right @yannmh I thought that when I sent the fix to the customer and then totally forgot.

yannmh:
Thanks @truthbk, let's merge it !

_:trophy:JMXFetch first merge achievement :trophy:_

PR #80, opened 2015-10-23
Support custom JMX urls

olivielpeau:
Allows using a non-custom URL for JMXFetch to connect to, on a per-instance basis.

See also https://github.com/DataDog/dd-agent/pull/1996, which allows adding custom JARs to JMXFetch's classpath.


yannmh:
Looks great ! :jack_o_lantern:

PR #81, opened 2015-10-28
Bean name matching logic: OR→AND

yannmh:
`matchBeanName` was incorrectly reporting `true` as soon as one
bean parameter in a configuration was matching and regardless of the
others.
To sum up, fix bean parameters matching logic: `OR`→`AND`


truthbk:
:+1:

yannmh:
Thanks for the review @truthbk. I am merging it.

PR #83, opened 2016-03-08
[jmx] if the exception is an IOException, throw it. Instance should be removed.

truthbk:
The IOException should be handled here:  

https://github.com/DataDog/jmxfetch/blob/master/src/main/java/org/datadog/jmxfetch/App.java#L204

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,6 @@ jdk:
 notifications:
   slack:
     secure: bUPdpDtrBYvlJ93AmfqvPU6ZSZhFHZJiIoNf0bXJVNZyxp2rjfMC8BG6XkesbKk62s7NJM0yfZhhLAzjNG3tSA9iL5i3npbl8kRaUi2zEDOCEG0zZ+FX2NxTiNH937S6QRmCBHe9/KEkSRuMY3HNRUidAq8w6dVCrWxxj5NwxjU=
+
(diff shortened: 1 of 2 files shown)


truthbk:
@yannmh this is supposed to address the case when one of the instances goes down. If we don't handle the exception in the correct place we apparently enter a messy loop that potentiallys ends up having the jmxfetch stop reporting metrics for a rather large window of time ~15+ minutes if I remember correctly.

.travis.yml
```diff
@@ -6,3 +6,6 @@ jdk:
 notifications:
   slack:
     secure: bUPdpDtrBYvlJ93AmfqvPU6ZSZhFHZJiIoNf0bXJVNZyxp2rjfMC8BG6XkesbKk62s7NJM0yfZhhLAzjNG3tSA9iL5i3npbl8kRaUi2zEDOCEG0zZ+FX2NxTiNH937S6QRmCBHe9/KEkSRuMY3HNRUidAq8w6dVCrWxxj5NwxjU=
+
+addons:
+  hostname: dd-jmxfetch-testhost
```
truthbk: addresses https://github.com/travis-ci/travis-ci/issues/5227

yannmh:
Nice catch @truthbk ! It looks good, feel free to merge it.

PR #85, opened 2016-03-21
[reporter][statsd] Allow configuration of statsd host

olivielpeau:
An optional host can be specified when the `statsd` reporter is used. This allows the Agent to make JMXFetch send its dogstatsd payloads to the `bind_host` configured in `datadog.conf`


yannmh:
Looks great. Feel free to merge it.

PR #89, opened 2016-04-05
[jmxattribute] remove `|` characters from tags

yannmh:
DogStatsd fails parsing statsd datagrams with `|` characters in the
metric tags.
Sanitize the metric tags: remove all `|` occurences in MBeans parameter
names and values.


olivielpeau:
Nice fix :+1: feel free to merge!

PR #93, opened 2016-04-15
[instance] Simplify metric match counting

JohnLZeller:
We have some weird metric counting going on. My initial solution here is to literally just use a simple count. This works well for `list_everything`. Gonna check everything else to make sure we are g2g tho.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -191,7 +191,7 @@ public class Instance {
 
         this.matchingAttributes.clear();
         this.failingAttributes.clear();
-        int metricsCount = 0;
+        int metricMatchCount = 0;
 
         if (!action.equals(AppConfig.ACTION_COLLECT)) {
             reporter.displayInstanceName(this);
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -252,14 +252,14 @@ private void getMatchingAttributes() {
                     try {
                         if (jmxAttribute.match(conf)) {
                             jmxAttribute.setMatchingConf(conf);
-                            metricsCount += jmxAttribute.getMetricsCount();
+                            metricMatchCount += 1;
```
yannmh: For `JMXComplexAttribute`s, the number of metric returned can be higher than `1`.

yannmh:
I don't think we are there yet: incrementing the counter is not a right approach (see my comment). 

I reckon it's also the reason why the metric counting behaves unexpectedly sometimes: the `list_*` commands might list `JMXComplexAttribute`s once only, when multiple fields are returned.

For example:
```
1/300 Fetched MyComplexAttribute  # 4 fields/metrics returned
5/300 Fetch  AnotherAttribute
``` 
I didn't have a chance to investigate so it's just an hypothesis.

Additionally, it would be nice to have some kind of tests to reproduce the issue, and emphasize the fix.

arbll:
Stale and would need work to get merged, closing

PR #79, opened 2015-10-20
Yann/cassandra metric mapping update

yannmh:
Rebase of https://github.com/DataDog/jmxfetch/pull/50

**Added changes**
Create a `cassandra_aliasing` flag to select the appropriate aliasing logic.
* `cassandra_aliasing` → compliant with CASSANDRA-4009
* default behavior

PR #94, opened 2016-04-18
Add option to generate alias from regexp match

alz:
This patch allows dynamic creation of metric aliases based on a regex match against <Bean>.<Attribute>. Example yaml with bean name "metrics:name=COUNTER1":

attribute:
            Count:
              metric_type: counter
              alias_match: metrics:name=(.*)
              alias: service.$1

Output alias becomes "services.counter1.count".

This greatly helps the creation of a generic config that splits hundreds (or thousands) of similarly named beans into different aliases, without the difficulty of grouping them under one alias and exceeding the maximum cardinality.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
@@ -342,6 +342,10 @@ public abstract class JMXAttribute {
         return attribute;
     }
 
+    public ObjectName getBeanName() {
+      return beanName;
+    }
+
     @SuppressWarnings("unchecked")
     protected String[] getTags() {
         if(tags != null) {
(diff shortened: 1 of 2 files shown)


ibra-islam:
+1

yannmh:
Thanks a lot for tackling the problem @alz.

Out of curiosity, could you let us know which of your Java application is causing these alias conflicts ?
We had to tackle quite a similar problem with Cassandra(>2.2) MBeans: multiple (=a LOT) of MBeans were being reported under the same name, but different tagging. We solved the problem by introducing a new `cassandra_aliasing` aliasing system (c.f. #50, #79) which instructs JMXFetch to remap metric names following the `cassandra.$name.$attribute` format rather than `cassandra.$attribute`. 

I just want to make sure that we are not solving twice the same problem :).

alz:
Hi @yannmh, we're monitoring one of our internal applications which exposes a very high number of metrics in in the form "metrics:type.<id>" with attributes such as "Count", etc. It's not feasible to add a config for each bean, and by selecting them with domain or bean regex matching it would group all attributes under the same alias and split the different beans into tags. It does sound similar to the Cassandra problem (except we're not using Cassandra) but gives more fine-grained control to how the alias is constructed.

yannmh:
Thanks for the extra explanations @alz.

Your changes look great! I rebased your PR on top of `master` to fix the tests and added some extra changes: let's move to #97.

PR #96, opened 2016-04-27
[service-checks] adds tags from configuration file to service checks

gmmeyer:
Currently, only the scope and instance tags are available on service checks. 

This PR will make it so that, in addition to those tags, all the tags defined in the configuration file will be available to service checks, making it much easier to scope an integration monitor to a set of hosts.

### DIFF
--- a/src/test/java/org/datadog/jmxfetch/TestServiceChecks.java
+++ b/src/test/java/org/datadog/jmxfetch/TestServiceChecks.java
@@ -43,7 +43,7 @@ public class TestServiceChecks extends TestCommon {
 
         assertEquals(Reporter.formatServiceCheckPrefix("jmx"), scName);
         assertEquals(Status.STATUS_OK, scStatus);
-        assertEquals(scTags.length, 1);
+        assertEquals(scTags.length, 3);
         assertTrue(Arrays.asList(scTags).contains("instance:jmx_test_instance"));
     }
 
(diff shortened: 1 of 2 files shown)


gmmeyer:
I had to make small changes to some of the tests. With this change they were expecting the wrong number of tags.

src/test/java/org/datadog/jmxfetch/TestServiceChecks.java
```diff
@@ -83,7 +83,7 @@ public void testServiceCheckWarning() throws Exception {
         assertEquals(Reporter.formatServiceCheckPrefix("too_many_metrics"), scName);
         // We should have an OK service check status when too many metrics are getting sent
         assertEquals(Status.STATUS_OK, scStatus);
-        assertEquals(scTags.length, 1);
+        assertEquals(scTags.length, 3);
         assertTrue(Arrays.asList(scTags).contains("instance:jmx_test_instance"));
```
yannmh: Can we tests the value of the 2 extra tags too ?

yannmh:
It looks good to me. 
I added a small nitpick, although it is ready to be merged.

src/test/java/org/datadog/jmxfetch/TestServiceChecks.java
gmmeyer: Yes, I can do that. In fact it was a good idea. The way I was adding tags made these tags:

```yml
env: stage
newTag: test
```
be output as just `stage` and `test` I've added another commit that will change that to `env:stage` `newTag:test` which is what I assume we want.

yannmh:
Nice, let's merge it.

PR #97, opened 2016-04-27
[jmxattribute] generate alias from regexp match

yannmh:
Rebase of #94.

**Additional changes**

* Refactor `getAlias` logic from `JMXSimpleAttribute` &
  `JMXComplexAttribute` to `JMXAttribute`
* Allow group name substitutions in `attribute`/`alias` parameter.
  Available options are:
  * `$attribute` for the attribute name
  * `$<parameter>` for any of the bean parameters
* Test coverage

Example:
```
bean: org.datadog.jmxfetch.test:foo=Bar,qux=Baz
attribute:
  toto:
    alias: my.metric.$foo.$attribute
```
returns a metric name `my.metric.bar.toto`.

Thanks again @alz !

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
@@ -4,13 +4,14 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.management.AttributeNotFoundException;
@@ -24,6 +25,8 @@ import org.apache.log4j.Logger;
 
 public abstract class JMXAttribute {
 
+    protected static final String ALIAS = "alias";
+    protected static final String METRIC_TYPE = "metric_type";
     private final static Logger LOGGER = Logger.getLogger(JMXAttribute.class.getName());
     private static final List<String> EXCLUDED_BEAN_PARAMS = Arrays.asList("domain", "domain_regex", "bean_name", "bean", "bean_regex", "attribute");
     private static final String FIRST_CAP_PATTERN = "(.)([A-Z][a-z]+)";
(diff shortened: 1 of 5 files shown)


src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -351,6 +355,89 @@ MBeanAttributeInfo getAttribute() {
...
+        Filter include = getMatchingConf().getInclude();
+        LinkedHashMap<String, Object> conf = getMatchingConf().getConf();
+
+        String fullAttributeName =(field!=null)?(getAttribute().getName() + "." + field):(getAttribute().getName());
+
+        if (include.getAttribute() instanceof LinkedHashMap<?, ?>) {
+            LinkedHashMap<String, LinkedHashMap<String, String>> attribute = (LinkedHashMap<String, LinkedHashMap<String, String>>) (include.getAttribute());
+            if (attribute.get(fullAttributeName).get("alias_match") != null) {
+              Pattern p = Pattern.compile(attribute.get(fullAttributeName).get("alias_match"));
+              Matcher m = p.matcher(getBeanName().getCanonicalName() + "." + fullAttributeName);
```
yannmh: @alz I am wondering if we should not match on the bean names only, i.e. drop the the attribute name. What do you think ?

alz: My thinking was that as the alias_match was defined inside the attribute yaml config it should include it in the match statement, and it also removed some extra duplicity where you would have to append ".#attribute#" to the end of the alias config line to reproduce the full metric name.

Of course using this method if you didn't want the same attribute name you'd have to add it to the alias_match regex, so it's a similar issue either way.

yannmh:
Thanks for your feedback @alz. 

On second thought, I think the`alias_match` parameter would be a little nebulous: from a user perspective, it's difficult to test and not clear to guess where the regular expression is applied (bean name versus attribute name for instance).

I went for another approach, which I think is easier to set up: allowing group name substitution directly into the `attribute`/`alias` parameter. The available options would be:
* ` $attribute` for the attribute name
* `$<parameter>` for any of the bean parameters

For example, the following configuration
```
bean: org.datadog.jmxfetch.test:foo=Bar,qux=Baz
attribute:
  toto:
    alias: my.metric.$foo.$attribute
```
would return a metric name `my.metric.bar.toto`.

Can you think of any use case where this approach might be lacking ? Thanks again.

alz:
That seems to make sense. If I'm understanding right a bean named "domain:name=foo,bar=baz" would have $name and $bar mapped? Might also be good if we could map $domain too that way all bases are covered.

yannmh:
That's right.

I added the `$domain` group name like you suggested. Let's merge it.

PR #102, opened 2016-06-03
Do not scope MBeans queries on `list_not_matching_attributes` action

olivielpeau:
Kudos to @nwillems for finding this bug! (https://github.com/DataDog/jmxfetch/commit/4d9b0fab97432c454ec4d9bc354d3dd056ef012f#commitcomment-17642270)

cc @yannmh

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -294,7 +294,7 @@ public class Instance {
     private void refreshBeansList() throws IOException {
         this.beans = new HashSet<ObjectName>();
         String action = appConfig.getAction();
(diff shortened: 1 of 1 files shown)


nwillems:
Thanks for fixing this. The code looks good to me. 

I guess this is not easily testable? I'm currently thinking about doing some more testing for this component.

Once again, thank you for taking the time to fix this.

yannmh:
Thanks @nwillems @olivielpeau. 

> I guess this is not easily testable?

One way I could think of, would be to mock the `connection.queryNames` and make assertions on the passed argument (i.e. the scope queried). Then, we could test different scenarios based on the user's configuration and the JMX command. The downside is that this would likely require an additional dependency to the project, something like http://mockito.org/.

Still, I think this should go out :shipit: 

Yann

PR #109, opened 2016-10-19
[service_discovery][rpc] implementing an RPC server

truthbk:
This PR introduces an RPC server into jmx so that we may submit YAML configs via RPC calls and thus enable service_discovery for JMX.

This is still a work in progress, but is functional at this stage. The code relies on [GRPC.io](http://www.grpc.io/) - server is on the java side, client side is in python in the dd-agent (responsible for the actual configuration management and service discovery.


truthbk:
We decided to go with named pipes, a lot more lightweight. Still good reference here if we ever implement an RPC server.

PR #104, opened 2016-08-05
Added JMX TabularData support

brothhaar:
Hi, this is a bit sloppy and there are no tests, but I wanted to get some feedback. Is this of any value to anyone else? 

* Added JMX `TabularData` support
* When a TabularData type is encountered, multiple metrics are generated with the same alias, but a unique tag per entry
* The tag is generated by concatenating each of the (possibly multiple) key values for each CompositeData entry in the TabularData JMX value
* Also included `limit` and `sort` capability to allow sorting and limiting the number of metrics returned for a single configuration entry (since a `TabularData` value could have many records).
* YAML config format:
```
      - include:
          domain: <domain>
          bean:
            - <bean name>
          attribute:
            <TabularData attribute name>:
              metric_type: <type>
              alias: <alias>
              tags:
                <tag name>: $key
              limit: 20
              sort: desc
```

PR #111, opened 2016-10-20
Added JMX TabularData support

brothhaar:
Copied from #104, since I screwed up that branch. 

Added JMX TabularData support
When a TabularData type is encountered, multiple metrics are generated with the same alias, but a unique tag per entry
The tag is generated by concatenating each of the (possibly multiple) key values for each CompositeData entry in the TabularData JMX value
Also included limit and sort capability to allow sorting and limiting the number of metrics returned for a single configuration entry (since a TabularData value could have many records).
YAML config format:

```
      - include:
          domain: <domain>
          bean:
            - <bean name>
          attribute:
            <TabularData attribute name>:
              metric_type: <type>
              alias: <alias>
              tags:
                <tag name>: $key
              limit: 20
              sort: desc
```

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -26,6 +26,7 @@ public class Instance {
             "java.util.concurrent.atomic.AtomicInteger", "java.util.concurrent.atomic.AtomicLong",
             "java.lang.Object", "java.lang.Boolean", "boolean", "java.lang.Number");
     private final static List<String> COMPOSED_TYPES = Arrays.asList("javax.management.openmbean.CompositeData", "java.util.HashMap");
+    private final static List<String> MULTI_TYPES = Arrays.asList("javax.management.openmbean.TabularData");
     private final static int MAX_RETURNED_METRICS = 350;
     private final static int DEFAULT_REFRESH_BEANS_PERIOD = 600;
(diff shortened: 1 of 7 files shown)


yannmh:
Thanks for the hard work @brothhaar ! 🙇 

Your changes are quite significant. We'll do our best to review, test it and follow up as soon as possible. As we currently are in the process of releasing Datadog Agent `5.10.0`, the goal is to have it out for the Datadog Agent `5.11.0` 🚀 .

brothhaar:
Thanks @yannmh. I still have one test failure I'm working out, but after that it should be good to go.

brothhaar:
OK @yannmh, the transient test failure I was seeing was related to counter metric rate calculation after a bean refresh (which I added to the test). The timing changed and I was seeing rates of `0.96 - 1.0` rather than the `0.98 - 1` that the `TestApp` was expecting. It's not a new bug, IMO, just odd timing in the test. I updated the metric's expected range.

Sorry about the delay on this, I'm just now cleaning up my TODO list from the past few months.

yannmh:
I apologize for the late review @brothhaar. 

The changes looks great! I forked your changes to add some slight modifications and opened a new PR: https://github.com/DataDog/jmxfetch/pull/128. I'd very appreciate to have your feedback on it before I merge it.

Thanks a lot again for the hard work!

PR #113, opened 2016-11-09
[service_discovery] enabling service discovery via named pipe.

truthbk:
### What does this PR do?

Enables service discovery for JMXfetch.

Once a service discovery config is detected by the agent it will be pushed as YAML via a named pipe: `dd-service_discovery` that lives in /tmp by default.

If multiple changes are detected, these will be submitted to JMXfetch in a single write operation to avoid race conditions. The pipe should take care of locking from the reading and writing ends, so we can guarantee all configs will be read on the JMXfetch in a single iteration and thus help avoid having multiple versions of the same config in the pipe at any given time. We will then do the chopping up of the various configurations read with a simple String.split(SD_CONFIGURATION_SEPARATOR) operaion.

### Motivation

Enabling Service Discovery for JMXfetch integrations.

### Testing Guidelines

An overview on testing
is available in our contribution guidelines.

### Additional Notes

Related PR: https://github.com/DataDog/dd-agent/pull/3010

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: java
 jdk:
+- oraclejdk8
 - oraclejdk7
 - openjdk7
 - openjdk6
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,7 @@
 		<jcommander.version>1.35</jcommander.version>
 		<junit.version>4.11</junit.version>
 		<log4j.version>1.2.17</log4j.version>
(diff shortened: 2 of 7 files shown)


hkaj:
Tests failed on a certificate issue on maven's repos, re-triggered them. Once they pass, SHIPIT.

truthbk:
No changes to the logic. Just added a couple simple tests. We're green, so merging!

PR #115, opened 2016-12-07
Adding 'histogram' metric type to JMXFetch.

hush-hush:
We now support 'histogram' type for metric (in addition of 'gauge' and 'counter').

This is a first step toward blacklisting tags (as different points may have the same tags set).

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
@@ -70,7 +70,9 @@ public abstract class Reporter {
             String[] tags = Arrays.asList((String[]) metric.get("tags")).toArray(new String[0]);
 
             // StatsD doesn't support rate metrics so we need to have our own aggregator to compute rates
-            if (!"gauge".equals(metricType)) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -70,7 +70,7 @@ public void sendMetrics(LinkedList<HashMap<String, Object>> metrics, String inst
             String[] tags = Arrays.asList((String[]) metric.get("tags")).toArray(new String[0]);
 
             // StatsD doesn't support rate metrics so we need to have our own aggregator to compute rates
-            if (!"gauge".equals(metricType)) {
+            if ("counter".equals(metricType)) {
```
yannmh: I am worried that this breaks backward compatibility with configuration that did not have `counter` properly spelt (`!"gauge".equals` versus `"counter".equals`).  What do you think?

hush-hush: Indeed ! I will reverse the condition to explicitly catch "gauge" and "histogram".
I guess that for the next major release, we should clarify the condition to only accept known types.

yannmh:
:shipit:

PR #116, opened 2016-12-07
Maxime/jmxfetch exclude tags

hush-hush:
This is usefull to filter out unique tags like host/client id, ...
This may/should be use in combination with the 'histogram' metric_type to avoid loosing metric points.


src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -73,6 +74,23 @@
...
+     */
+    private void applyTagsBlackList() {
+        Filter include = this.matchingConf.getInclude();
+        if (include != null) {
+
+            for (String excludedTagName : include.getExcludeTags()) {
+                for (Iterator<String> it = this.defaultTagsList.iterator(); it.hasNext();) {
+                    if (it.next().startsWith(excludedTagName + ":")) {
+                        it.remove();
+                    }
```
yannmh: Instead of this double `for` loop, can we have the `getExcludeTags` method to return a `Set`, and only iterates over `defaultTagsList ` performing lookups?

Issue #108, opened 2016-10-13
Improved tagging capabilities


krjackso:
I have a similar request  to #103 where I want to whitelist tags, specifically the `name` tag that comes from the MBean name property. 

The issue I'm having is that we have tagged our ec2 hosts with `name` and we can not easily get system-wide results when creating graphs with something along the lines of `get jmx.domain.attribute1, from attribute_group:group1, sum by name` since it splits it by both the bean name and the instance name. 

To solve this, you could use something similar to what you have done aliases e.g.

```yaml
include:
  attribute:
    attribute1:
      alias: jmx.$domain.$attribute
      metric_type: gauge
      tags_include:
        - "attribute_group:group1"
        - "bean_name:$name"
  bean_regex: Domain:name=group1-*
```

I was also thinking though that what I really want is to put the tags on all attributes of the bean, not just one. The above config becomes repetitive with multiple attributes. It would be great if the config could be simplified further to:

```yaml
include:
  attribute:
    - attribute1
    - attribute2
  bean_regex: Domain:name=group1-*
  tags_include:
     attribute_group: group1
     bean_name: $name
```

Maybe I am missing something? Is there a way to achieve this with the current jmxfetch?

yannmh:
Thanks for the details @krjackso,

This issue actually falls within a bigger scope: the lack of name-spacing in tags.  Integrations produce various tags, with more or less generic names, which can conflicts with the host tags. 
We had a similar issue before, with JMXFetch producing a `host` tag. This constrained us to rename it to `jmx_server`, c.f. https://github.com/DataDog/jmxfetch/blob/master/CHANGELOG.md#080--09-17-2015

While we are working on resolving this problem at a higher level (for all integrations), we think it would make sense to implement your proposal with #103.

krjackso:
Updating this with additional information... A use case that my organization (Tealium) is especially interested in, is being able to capture the regular expression groups in the bean name and use those in the enhanced tags.

From the jmx list_everything command:
```
> service datadog-agent jmx list_everything | grep ItemsProcessed
Not Matching: Bean name: Metrics:name=ItemsProcessed-account1-profile1 - Attribute name: Count  - Attribute type: long
Not Matching: Bean name: Metrics:name=ItemsProcessed-account1-profile1 - Attribute name: Mean  - Attribute type: double
Not Matching: Bean name: Metrics:name=ItemsProcessed-account1-profile2 - Attribute name: Count  - Attribute type: long
[...]

PR #117, opened 2016-12-09
Maxime/add additional tags option

hush-hush:
Add 'additional_tags' list to the configuration: ref #108

This allows a user to define additional tag at the 'include' level (with alias using the '$tag_name' notation).


yannmh:
Can we name it `tags` instead of `additional_tags` ? I realized this is how we name it in most of our checks shipped with the Datadog Agent.

hush-hush:
We already have a 'tags' section in the `instances` section and aliasing is not handled in that one. Do we mind having the same name for 2 sections ?

yannmh:
I think it makes sense: both fields represent additional tags, at different levels (instance versus attribute).

PR #119, opened 2017-01-13
Use https://oss.sonatype.org/content/groups/public/ instead of http:/…

dbaxa:
…/oss.sonatype.org/content/groups/public/ .

Signed-off-by: David Black <dblack@atlassian.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,14 +26,6 @@
 		<snakeyaml.version>1.13</snakeyaml.version>
 	</properties>
 
-	<repositories>
-		<repository>
-			<id>Sonatype-public</id>
-			<name>SnakeYAML repository</name>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -30,7 +30,7 @@
 		<repository>
 			<id>Sonatype-public</id>
 			<name>SnakeYAML repository</name>
-			<url>http://oss.sonatype.org/content/groups/public/</url>
+			<url>https://oss.sonatype.org/content/groups/public/</url>
```
dbaxa: I am not actually use that this repository is used.

yannmh: You are right, this can be completely removed. Would you like to add a commit for that? 🙇

dbaxa: Done.

yannmh:
Thanks @dbaxa 🙇

PR #122, opened 2017-01-31
Print exception when unknown IOException occured

aoking:
The IOExceotion's stacktrace is not printed when using JMX Attachi API.
This behavior is too inconvenient.


olivielpeau:
Thanks @aoking! Makes a lot of sense, merging

PR #124, opened 2017-02-23
Fix JMXFetch issue when several instances on the same host

zippolyte:
### What this PR does
It fixes the issue occurring when refreshing shared connections by creating one connection per instance and thus removing shared connection.

### Motivation
In the connection manager, it was possible to get a connection shared between different instances. 
If at some point the connection gets broken or the App needs to refresh, it will force a new connection for every instance sharing this connection. For the first instance it will work properly and create a fresh connection, then for the second instance, it will not use the freshly created connection but close it and recreate a new one. Doing so, it breaks the connection of the first instance which will in turn force a new connection being open, breaking the one of the first instance, etc...., entering a loop.

My solution is to avoid sharing connections and create one for each instance even if they connect to the same host.


olivielpeau:
Thanks @zippolyte for the fix!

I think the `instanceName` is not guaranteed to be unique across instances if 2 instances with the same host/port are defined in the same configuration file for example.

Also, even if the `instanceName` is unique, creating one connection per configured instance kind of defeats the purpose of the `ConnectionManager` class:

> Singleton used to share connections across instances in case you have multiple instances configured to the same MBeanServer (e.g. Solr and Tomcat)

(see https://github.com/DataDog/jmxfetch/blob/0.12.0/src/main/java/org/datadog/jmxfetch/ConnectionManager.java#L11)

Instead of this solution, have you explored if we could keep sharing connections between instances and fix the logic that handles the connection refresh (so that it only refreshes the shared connection once and makes all the instances that use this shared connection use it)?

I'm not sure such a fix would be easy to implement, so if it doesn't look worth it we may want to keep your current solution. In that case we could improve this fix by ensuring that the instanceName is unique, and we should probably update the `ConnectionManager` class so that its only responsibility is creating a connection.

zippolyte:
After some thinking, keeping the shared connections did not seem worth the trouble. Indeed it would imply making a lot more calls to the `ConnectionManager` to make sure the last version of the connection is used (it might have been refreshed by another instance) and passing this up-to-date connection to a lot of different objects methods (e.g. `JMXAttribute.getValue`). And the performance gained with these shared connections would be negligible since there are usually only two instances sharing a connection.
So it was decided that each instance would keep its own connection and the `ConnectionManager` would become a `ConnectionFactory` with the only purpose of creating connections.

PR #126, opened 2017-02-24
[jmxfetch][pom] include build commit in manifest.

truthbk:


pom.xml
```diff
@@ -118,19 +118,49 @@
...
+                <executions>
+                    <execution>
+                        <phase>validate</phase>
+                        <goals>
+                            <goal>create</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <doCheck>false</doCheck>
```
truthbk: If we set this to `true` the build will fail on uncommited changes. I think it's better at `false` for contributors, but we can change it.

yannmh:
Looks good to me. Let's :shipit:

PR #128, opened 2017-03-01
[attribute] add TabularData support

yannmh:
Original discussion thread: #111
Credits to @brothhaar. Thanks a lot !!! 🙇 

**JMX TabularData support**

_Copied from  #111, @brothhaar._

When a TabularData type is encountered, multiple metrics are generated with the same alias, but a unique tag per entry.
The tag is generated by concatenating each of the (possibly multiple) key values for each CompositeData entry in the TabularData JMX value
Also included limit and sort capability to allow sorting and limiting the number of metrics returned for a single configuration entry (since a TabularData value could have many records).
YAML config format:
```yaml
- include:
  domain: <domain>
  bean:
    - <bean name>
  attribute:
    <TabularData attribute name>:
      metric_type: <type>
      alias: <alias>
      tags:
        <tag name>: $key
      limit: 20
      sort: desc
```

## Additional changes

**Allow tag substitution for any key of a `TabularData` object**

Do not restrain tag substitution to the `TabularType` index. Instead, tag subtitution can be performed on any keys of a `CompositeType`.

For example:
```yaml
Tabulardata.foo:
    metric_type: gauge
    alias: multiattr.foo
    tags:
      foo: $foo
      toto: $toto
    limit: 1
    sort: desc
```
instruments JMXFetch to collect a `multiattr.foo` metric tagged with `foo` and `toto` keys, where `foo` and `toto` are `CompositeType` keys. 

**Others**
* Modify tests to use `sort` and `limit` for `TabularData`
* Rename `JMXMultiAttribute` to `JMXTabularAttribute`

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -249,11 +249,11 @@ public class Instance {
                     LOGGER.debug(ATTRIBUTE + beanName + " : " + attributeInfo + " has attributeInfo simple type");
                     jmxAttribute = new JMXSimpleAttribute(attributeInfo, beanName, instanceName, connection, tags, cassandraAliasing);
                 } else if (COMPOSED_TYPES.contains(attributeType)) {
-                    LOGGER.debug(ATTRIBUTE + beanName + " : " + attributeInfo + " has attributeInfo complex type");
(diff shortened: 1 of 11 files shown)


brothhaar:
👍 Looks good @yannmh. Thanks for taking the time to pull this in!

PR #130, opened 2017-03-02
[attribute][complex] extend to `java.util.Map`

### LINKED ISSUE #125: No support for type: java.util.map
Migrating a bunch of hosts from current monitoring solution to Datadog, discovered that this JMX type was not supported by Datadog.

Example output from JTerm:
```
#mbean = sme:name=ActiveCampaignStats:
CampaignCount = {
  Europe = 4563;
  Asia = 1128;
  All = 82134;
 };
$>
```

Which in JConsole might look something like:
```
AttributeName:        AttributeValue:
=======================================
CampaignCount         {Europe=4563,Asia=1128,All=82134}
```

yannmh:
Add `java.util.Map` to the list of supported `JMXComplexAttribute`
types.

Fix #125

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/JMXComplexAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXComplexAttribute.java
@@ -34,8 +34,8 @@ public class JMXComplexAttribute extends JMXAttribute {
             for (String key : data.getCompositeType().keySet()) {
                 this.subAttributeList.put(key, new HashMap<String, Object>());
             }
-        } else if ("java.util.HashMap".equals(attributeType)) {
-            HashMap<String, Double> data = (HashMap<String, Double>) attributeValue;
+        } else if (("java.util.HashMap".equals(attributeType)) || ("java.util.Map".equals(attributeType))){
+            Map<String, Double> data = (Map<String, Double>) attributeValue;
             for (String key : data.keySet()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jmxfetch/JMXComplexAttribute.java
```diff
@@ -34,7 +34,7 @@ private void populateSubAttributeList(Object attributeValue) {
             for (String key : data.getCompositeType().keySet()) {
                 this.subAttributeList.put(key, new HashMap<String, Object>());
             }
-        } else if ("java.util.HashMap".equals(attributeType)) {
+        } else if (("java.util.HashMap".equals(attributeType)) || ("java.util.Map".equals(attributeType))){
             HashMap<String, Double> data = (HashMap<String, Double>) attributeValue;
```
olivielpeau: I don't think this cast will work in the general case when the underlying `attributeValue` object is not a HashMap.

Have you tested this with a non-`HashMap` `Map`?

olivielpeau: (we could probably just cast `attributeValue` to a `Map` right?)

same applies to your change line 86

yannmh: You're right it should stay a `Map`

Issue #131, opened 2017-03-03
Error parsing list of tags


masci:
JMX-fetch only expects tags to be listed as a map in the YAML file but there are cases when they might be arrays. One example is the output of the template parsing in Service Discovery, see https://github.com/DataDog/dd-agent/blob/69d58afeea2126ab2a7e1e1efba6f3cbdafc7f33/utils/service_discovery/sd_docker_backend.py#L410-#L420

In such cases the error looks like this:
```
java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.util.LinkedHashMap
	at org.datadog.jmxfetch.Instance.<init>(Instance.java:71)
	at org.datadog.jmxfetch.App.init(App.java:485)
	at org.datadog.jmxfetch.App.start(App.java:241)
	at org.datadog.jmxfetch.App.main(App.java:122)
```

We should fix JMX-fetch so it can accept tags in both formats.

yannmh:
Thanks for the details @masci. A fix is coming.

PR #132, opened 2017-03-03
[instance] support instance tags in list format

yannmh:
Support instance tags defined in an array/list format.

Fix https://github.com/DataDog/jmxfetch/issues/131


src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -135,7 +135,13 @@ private void addAdditionalTags() {
 
         if (instanceTags != null) {
             for (Map.Entry<String, String> tag : instanceTags.entrySet()) {
-                beanTags.add(tag.getKey() + ":" + tag.getValue());
+                if (tag.getValue() != null) {
+                    beanTags.add(tag.getKey() + ":" + tag.getValue());
+
```
olivielpeau: could you remove the extra blank line here?

Issue #123, opened 2017-02-17
NullPointerException if no attribute alias is specified


sbarlabanov:
The following attribute configuration results in a NullPointerException in 0.12.0:

```
            - include:
               bean: org.datadog.jmxfetch.test:type=SimpleTestJavaApp,scope=Co|olScope,host=localhost,component=
               attribute:
                    ShouldBe100:
                        metric_type: gauge
```
I reproduced it in a unit test:
Exception is:
```
java.lang.NullPointerException
	at org.datadog.jmxfetch.JMXAttribute.getUserAlias(JMXAttribute.java:445)
	at org.datadog.jmxfetch.JMXAttribute.getAlias(JMXAttribute.java:385)
	at org.datadog.jmxfetch.JMXAttribute.getAlias(JMXAttribute.java:462)
	at org.datadog.jmxfetch.JMXSimpleAttribute.getMetrics(JMXSimpleAttribute.java:31)
	at org.datadog.jmxfetch.Instance.getMetrics(Instance.java:164)
	at org.datadog.jmxfetch.App.doIteration(App.java:182)
	at org.datadog.jmxfetch.TestCommon.run(TestCommon.java:99)
	at org.datadog.jmxfetch.TestApp.testBeanTags(TestApp.java:31)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
[...]

PR #133, opened 2017-03-03
[attribute][conf] Allow specifying no alias on detailed attribute

olivielpeau:
Fixes https://github.com/DataDog/jmxfetch/issues/123

We simply pick the next correct default if no alias is provided


yannmh:
Looks good. Thank you.

PR #135, opened 2017-04-17
[service_discovery] Fix race condition preventing SD initialization

olivielpeau:
If JMXFetch starts before the collector process creates the SD pipe,
JMXFetch doesn't find the SD pipe and never re-tries opening it (until
it's fully restarted).

To solve this, retry opening the SD pipe at every run iteration when
SD is enabled. Also, add an `--sd-enabled` option to know whether
JMXFetch should enable SD (i.e. whether it should expect the SD pipe
to exist or not).

The `--sd-standby` option is not used and has therefore been removed.

This change requires a change to `dd-agent` to make `jmxfetch.py` pass
the `--sd-enabled` option when SD-JMXFetch is enabled. See https://github.com/DataDog/dd-agent/pull/3306

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -237,8 +237,8 @@ public class App {
                 sdPipe = newSdPipe();
             }
             try {
-                if(sdPipe != null && sdPipe.available() > 0) {
-                    int len = sdPipe.available();
+                int len;
+                if(sdPipe != null && (len = sdPipe.available()) > 0) {
                     byte[] buffer = new byte[len];
                     sdPipe.read(buffer);
                     setReinit(processServiceDiscovery(buffer));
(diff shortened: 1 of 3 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -225,13 +232,17 @@ void start() {
             }
 
             // any SD configs waiting in pipe?
+            if(sdPipe == null && appConfig.getSDEnabled()) {
+                // If SD is enabled and the pipe is not open, retry opening pipe
+                sdPipe = newSdPipe();
+            }
             try {
```
derekwbrown: Little nit, but could do
int len = 0
if(sdPipe != null && ((len = sdPipe.available()) > 0)) {
   byte[] buffer....

No need to call sdPipe.available twice in succession.

olivielpeau: 👍 , addressed in 1c5bede5566b07788e52b99a92e16372d8f290d1

truthbk: Nice, one. My bad.

truthbk:
There's a weird Travis error for oraclejdk7, let's make sure we know what's up with that before merging. I think it's unrelated.

olivielpeau:
Must've been a flaky test, I've just re-run the tests on oraclejdk7 and they're all green now!

Merging

PR #138, opened 2017-04-27
adds min collection interval

gmmeyer:
We had a request for this change: someone wanted to be able to configure the collection period for each jmx check. This will allow you to do that.

Right now the only thing you can do is to configure the loop for all checks, this allows you to configure the loop for each check, or even for each instance of each check.


src/test/java/org/datadog/jmxfetch/TestInstance.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.LinkedList;
+
+import org.apache.log4j.Logger;
+import org.junit.Test;
+
+public class TestInstance extends TestCommon {
+	private final static Logger LOGGER = Logger.getLogger("Test Instance");
+	
+	@Test
+	public void testMinCollectionInterval() throws Exception {
```
masci: ❤️

PR #140, opened 2017-05-31
[jmxfetch] option should be `min_collection_interval` in the YAML.

truthbk:


olivielpeau:
I think there's a test that needs an update too:
https://github.com/DataDog/jmxfetch/blob/7cf612c454088f7bf758d6bed6f99550fd84b5a9/src/test/java/org/datadog/jmxfetch/TestInstance.java and https://github.com/DataDog/jmxfetch/blob/7cf612c454088f7bf758d6bed6f99550fd84b5a9/src/test/resources/jmx_min_collection_period.yml ;)

gmmeyer:
@olivielpeau ah ha you're right!

truthbk:
Merging not that CI is back to 💚

PR #145, opened 2017-07-10
[auto_discovery] process templates larger than the page buffer size

truthbk:
We must be able to process templates as large as necessary - customer encountered we were just processing 64K configurations at most when reading from the pipe - consistent with the default page buffer on linux. Let's read the entire buffer... Also, this is necessary to avoid blocking the writing side indefinitely.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -249,7 +249,7 @@ public class App {
                 adPipe = newAutoDiscoveryPipe();
             }
             try {
-                if(adPipe != null) {
+                if(adPipe != null && adPipe.available() > 0) {
                     byte[] buffer = new byte[0];
                     boolean terminated = false;
                     while (!terminated) {
@@ -261,7 +261,7 @@ public class App {
                             // The separator always comes in its own atomic write() from the agent side -
                             // so it will never be chopped.
                             if (Bytes.indexOf(minibuff, App.AD_LEGACY_CONFIG_TERM.getBytes()) > -1 ||
-                                Bytes.indexOf(minibuff, App.AD_CONFIG_TERM.getBytes()) > -1 ) {
+                                    Bytes.indexOf(minibuff, App.AD_CONFIG_TERM.getBytes()) > -1 ) {
                                 terminated = true;
                             }
 
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -171,12 +171,12 @@ public class App {
 
     private FileInputStream newAutoDiscoveryPipe() {
         FileInputStream adPipe = null;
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -244,14 +245,25 @@ void start() {
...
             }
             try {
-                int len;
-                if(adPipe != null && (len = adPipe.available()) > 0) {
-                    byte[] buffer = new byte[len];
-                    adPipe.read(buffer);
+                if(adPipe != null) {
+                    int offset = 0;
+                    byte[] buffer = new byte[0];
+                    while (adPipe.available() > 0) {
```
theduderog: Is there a possible race condition here where `adPipe.available()` might return 0 after the read but before the writer is unblocked?

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -244,14 +245,25 @@ void start() {
                 adPipe = newAutoDiscoveryPipe();
             }
             try {
-                int len;
-                if(adPipe != null && (len = adPipe.available()) > 0) {
-                    byte[] buffer = new byte[len];
-                    adPipe.read(buffer);
+                if(adPipe != null) {
+                    int offset = 0;
```
olivielpeau: you can remove this var

src/main/java/org/datadog/jmxfetch/App.java
olivielpeau: hmm good point, what do you think @truthbk?

truthbk: I'd have to look at the internals, I'm not sure. I haven't encountered issues while testing, but this is a valid point.

truthbk:
After taking a look at the internals, I believe @theduderog was indeed correct and we will need to use a separator to guarantee we know when no more data is expected in the iteration. Will have to fix that here, and in the `dd-agent` side.

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -41,6 +43,8 @@
     public static final String CANNOT_CONNECT_TO_INSTANCE = "Cannot connect to instance ";
     private static final String AD_CONFIG_SEP = "#### AUTO-DISCOVERY ####";
     private static final String AD_LEGACY_CONFIG_SEP = "#### SERVICE-DISCOVERY ####";
+    private static final String AD_CONFIG_TERM = "#### AUTO-DISCOVERY TERM ####";
+    private static final String AD_LEGACY_CONFIG_TERM = "#### SERVICE-DISCOVERY TERM ####";
```
hush-hush: The end header is `"#### SERVICE-DISCOVERY END ####\n"` and not `#### SERVICE-DISCOVERY TERM ####`.
from: https://github.com/DataDog/dd-agent/pull/3426/files#diff-05b1a4892667b4d4a36c7ac84208c8c0R77

PR #147, opened 2017-07-18
Maximum length for instance configs should be longer - improve tests.

truthbk:


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -415,7 +425,16 @@ public void doIteration() {
 
     public boolean addConfig(String name, YamlParser config) {
         // named groups not supported with Java6: "(?<check>.{1,30})_(?<version>\\d{0,30})"
```
xvello: Should that comment be updated?

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -415,7 +425,16 @@ public void doIteration() {
...
+            LOGGER.debug("Name too long - skipping: " + name);
+            return false;
+        }
+        String patternText = AUTO_DISCOVERY_PREFIX+"_(.{1," + AD_MAX_NAME_LEN +
+            "})_(\\d{0,"+ AD_MAX_MAG_INSTANCES +"})";
+
+        Pattern pattern = Pattern.compile(patternText);
 
         Matcher matcher = pattern.matcher(name);
         if (!matcher.find()) {
```
xvello: I'd love to have debug output here too, in case it were to happen

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -45,6 +46,9 @@
     private static final String AD_LEGACY_CONFIG_SEP = "#### SERVICE-DISCOVERY ####";
     private static final String AD_CONFIG_TERM = "#### AUTO-DISCOVERY TERM ####";
     private static final String AD_LEGACY_CONFIG_TERM = "#### SERVICE-DISCOVERY TERM ####";
+    private static final int AD_MAX_NAME_LEN = 80;
+    private static final int AD_MAX_MAG_INSTANCES = (int)Math.log10(1000) + 1;
```
xvello: That's a bit cryptic, I'd go with `4  // 1000 instances ought to be enough for anyone`

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -414,19 +424,30 @@ public void doIteration() {
     }
 
     public boolean addConfig(String name, YamlParser config) {
-        // named groups not supported with Java6: "(?<check>.{1,30})_(?<version>\\d{0,30})"
-        Pattern pattern = Pattern.compile(AUTO_DISCOVERY_PREFIX+"(.{1,30})_(\\d{0,30})");
+        // named groups not supported with Java6: "(?<check>.{1,80})_(?<version>\\d{0,80})"
```
xvello: I think it's `(?<version>\\d{0,4})` right, using `AD_MAX_MAG_INSTANCES`?

PR #149, opened 2017-07-22
Support value conversion for complex types

tinexw:
It's currently not possible to define value conversions for map entries e.g.

    somemap.field1:
      values:
        default: 42

This PR adds support for it.


jeffwidman:
nudge @masci can you get someone internally to take a look at this?

PR #150, opened 2017-07-31
adds backoff

gmmeyer:
When there is a connection failure, we should not immediately declare the app broken. Instead, we should retry it with some backoff, only adding it to the list of the broken after it has failed four times

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -21,6 +21,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Random;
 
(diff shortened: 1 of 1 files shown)


olivielpeau:
Going to close this PR:

We don't actually need to implement a backoff mechanism: if a connection fails during a run of JMXFetch, it'll already be retried on the next run (i.e. 10 seconds later). The issue that was reported was that when the JVM JMXFetch connects to doesn't expose all of its beans _yet_ (for instance because that JVM has just started), since the connection is successful JMXFetch will wait a default period of 10 minutes before refreshing the beans: this makes JMXFetch miss all the beans that weren't exposed by the JVM at the first run for 10 minutes.

#151 addresses this issue.

gmmeyer:
thanks for closing it! I had forgotten to do that!

PR #151, opened 2017-08-10
set a minimum number of metrics to collect

gmmeyer:
If JMXfetch is unable to collect a minimum number of metrics from an app, it will refresh the beans at the beginning of the next run, rather than relying upon the number that it has cached. 

This is in response to a number of people who have found that it is annoying to have to bump the refresh period just to handle issues when the app being monitored is restarting.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -237,7 +237,7 @@ public class Instance {
             }
         }
         if (this.minReturnedMetrics < metrics.size()) {
-        	  this.refreshNextRun = true;
+        	this.refreshNextRun = true;
         }
         return metrics;
     }
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -237,7 +237,7 @@ public class Instance {
             }
         }
         if (this.minReturnedMetrics < metrics.size()) {
-        		this.refreshNextRun = true;
+        	    this.refreshNextRun = true;
         }
         return metrics;
     }
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -87,10 +87,8 @@ public class Instance {
 
         this.minCollectionPeriod = (Integer) yaml.get("min_collection_interval");
         if (this.minCollectionPeriod == null && initConfig != null) {
-        		this.minCollectionPeriod = (Integer) initConfig.get("min_collection_interval");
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -228,6 +236,9 @@ public String toString() {
                 }
             }
         }
+        if (this.minReturnedMetrics < metrics.size()) {
+        		this.refreshNextRun = true;
```
olivielpeau: could you fix the indentation here?

gmmeyer: the default eclipse indentation is weird

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -52,6 +52,8 @@
     private Connection connection;
     private AppConfig appConfig;
     private Boolean cassandraAliasing;
+    private Integer minReturnedMetrics = 0;
```
truthbk: I don't think we can modify this during runtime, should we make the attribute `final` (and set in the constructor like we're currently doing)? If we do, we'll have to handle the case when the option is not set (`Integer.MAX_VALUE` or `-1`?).

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -97,6 +100,11 @@ public Instance(LinkedHashMap<String, Object> yamlInstance, LinkedHashMap<String
             this.maxReturnedMetrics = (Integer) maxReturnedMetrics;
         }
 
+        Object minReturnedMetrics = this.yaml.get("min_returned_metrics");
```
truthbk: Can we document this in the configuration YAML?

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -228,6 +236,9 @@ public String toString() {
                 }
             }
         }
+        if (this.minReturnedMetrics < metrics.size()) {
```
truthbk: If we decide to make `minReturnedMetrics` `final` then we will have to initialize it to an integer (we could set a `-1` to reflect the option is disabled) in the constructor. If we don't make it `final` then we could check it's null before making this comparison or else I think we could run into an exception here.

src/main/java/org/datadog/jmxfetch/Instance.java
gmmeyer: yes

src/main/java/org/datadog/jmxfetch/Instance.java
gmmeyer: ah yea, this is a problem

PR #154, opened 2017-08-24
[rate] add canonical_rate + feature flag for the feature

truthbk:
This PR adds a feature flag to enable canonical rate behavior for counters/rates such that a negative rate would not be reported. A legit negative rate would typically represent a metric that _should_ be reported as a `gauge`. Because the former inconsistent behavior has been present in jmxfetch for a while, and to avoid breaking alerts and dashboards, we will put this behind a feature flag for now.

Based off #146

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
@@ -35,7 +35,7 @@ public abstract class Reporter {
         ratesAggregator.put(instanceName, new HashMap<String, HashMap<String, Object>>());
     }
 
-    public void sendMetrics(LinkedList<HashMap<String, Object>> metrics, String instanceName) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -88,7 +88,10 @@ public void sendMetrics(LinkedList<HashMap<String, Object>> metrics, String inst
                 long now = System.currentTimeMillis();
                 double rate = 1000 * (currentValue - oldValue) / (now - oldTs);
 
-                if (!Double.isNaN(rate) && !Double.isInfinite(rate)) {
+                boolean submit = (!Double.isNaN(rate) && !Double.isInfinite(rate));
+                submit &= (rate >= 0 | !canonicalRate);
+
+                if  (submit) {
                     sendMetricPoint(metricType, metricName, rate, tags);
                 }
```
olivielpeau: nit: I think we should log something when we don't submit a metric because `canonicalRate` is enabled, could ease troubleshooting missing rates.

PR #155, opened 2017-09-13
[agent6] add support for submission of JMX statuses to agent6.

truthbk:
Submits JMX status to Agent6. Implements backward compatibility with agent5.

## Note
We currently use self-signed certificates on agent6, so we're actually skipping certificate verification on the client. This will be removed once we get that covered.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Status.java
+++ b/src/main/java/org/datadog/jmxfetch/Status.java
@@ -140,9 +140,8 @@ public class Status {
 
             //add reuqest header
             con.setRequestMethod("POST");
-            //con.setRequestProperty("User-Agent", USER_AGENT);
             con.setRequestProperty("Content-Type", "application/json");
-            con.setRequestProperty("Session-Token", this.token);
+            con.setRequestProperty("Authorization", "Bearer "+ this.token);
 
             con.setDoOutput(true);
             DataOutputStream wr = new DataOutputStream(con.getOutputStream());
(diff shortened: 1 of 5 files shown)


src/main/java/org/datadog/jmxfetch/Status.java
```diff
@@ -87,15 +122,62 @@ private String generateYaml() {
...
+        int responseCode = 0;
+        try {
+            String url = "https://localhost:" + port + "/agent/jmxstatus";
+
+            URL uri = new URL(url);
+            HttpsURLConnection con = (HttpsURLConnection) uri.openConnection();
+
+            //add reuqest header
+            con.setRequestMethod("POST");
+            //con.setRequestProperty("User-Agent", USER_AGENT);
```
olivielpeau: let's remove this line unless you want to set this header?

src/main/java/org/datadog/jmxfetch/Status.java
```diff
@@ -87,15 +122,62 @@ private String generateYaml() {
...
+        HashMap<String, Object> status = new HashMap<String, Object>();
+        status.put("timestamp", System.currentTimeMillis());
+        status.put("checks", this.instanceStats);
+        return gson.toJson(status);
+    }
+
+    private boolean postRequest(String body, int port) {
+        int responseCode = 0;
+        try {
+            String url = "https://localhost:" + port + "/agent/jmxstatus";
```
olivielpeau: We make the assumption in a lot of places in the new agent that `localhost` actually works, but we've experienced a lot of cases where `localhost` doesn't work properly (and users have to customize this to `127.0.0.1` or `::`). The Statsd reporter actually supports the configuration of the host it reports to because of this.

Not something we need to tackle now, but we should do it at some point :)

PR #156, opened 2017-09-22
[api] collect instance configurations via API

truthbk:
This PR allows collecting configurations via the API potentially exposed by the agent. 

`Gson` is great, but I encountered some issues with the unmarshaling of the JSON, so I eventually switched it out for `Jackson` which seemed to do the trick just fine with a simpler interface.

This PR obsoletes the collection of configurations via named pipes - which was never ideal. The code however has not been removed for legacy reasons, and in fact both options can coexist as sources for the configurations. 

To better handle API calls, an `HttpClient` and `HttpClient.HttpResponse` classes have been introduced. Do not expect `HttpClient` to be entirely generic as it has been designed bearing in mind it's a client for our particular API - more specifically it encapsulates the "authentication" which would be absolutely meaningless on a different API.

Refactored a few vars and methods to avoid confusion.

Testing could be perhaps extended and or improved, but a basic case to address the JSON unmarshalling, etc has been added.


truthbk:
Tests broken on JDK6 - will look into it.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -42,7 +42,7 @@
     private long lastCollectionTime;
     private Integer minCollectionPeriod;
     private long lastRefreshTime;
-    private LinkedHashMap<String, Object> yaml;
+    private LinkedHashMap<String, Object> instanceMap;
```
gmmeyer: 🍰

PR #167, opened 2018-03-02
Adds support for bean regex tagging

nmuesch:
Adds support for using the bean_regex syntax to capture groups for tags. 

We currently support pulling in common attributes out of bean names, but this introduces the ability to specify any custom substring of a bean name to utilize as a tag on the associated attributes. 

Ex:

```
init_config:
  conf:
    - include:
	bean_regex: 'Catalina:type=(.*?),*host=(.*?),context=(.*),*'
        attribute:
          spareNotFoundEntries:
            metric_type: gauge
            alias: tomcat.spareNotFound
          cacheMaxSize:
            metric_type: gauge
            alias: tomcat.cacheMax
        tags:
             TypeRegex: $1
             HostRegex: $2
             contextRegex: $3
             optional: tag
```

This will submit the metric `tomcat.spareNotFound` with tags `hostRegex:<beanParameter>` ,`TypeRegex:<beanParameter>`, and `contextRegex<beanParameter>` as well as the tag with key `optional` and value `tag`

If group 3 in this example didn't exist, this logs a warning and doesn't apply the `contextRegex` tag to the metric.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
@@ -100,8 +100,8 @@ public abstract class JMXAttribute {
         if (include != null) {
             for (Map.Entry<String, String> tag : include.getAdditionalTags().entrySet()) {
             		String alias = this.replaceByAlias(tag.getValue());
-            		if (alias != "" || alias != null) {
-            			this.defaultTagsList.add(tag.getKey() + ":" + this.replaceByAlias(tag.getValue()));
+            		if ((alias.trim().length() > 0) && alias != null) {
+            			this.defaultTagsList.add(tag.getKey() + ":" + alias);
             		} else {
                     LOGGER.warn("Unable to apply tag " + tag.getKey() + " - with unknown alias");
             		}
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -295,7 +296,12 @@ private boolean matchBeanRegex(Filter filter, boolean matchIfNoRegex) {
         }
 
         for (Pattern beanRegex : beanRegexes) {
-            if(beanRegex.matcher(beanStringName).matches()) {
+            Matcher m = beanRegex.matcher(beanStringName);
+
+            if(m.matches()) {
+            	for (int i = 0; i<= m.groupCount(); i++) { 
+            		this.beanParameters.put(""+i, m.group(i));
```
gmmeyer: This is a stylistic issue, not a technical one. I don't love 
```java
""+i
```
I don't think it's very clear. I think it would be better to make it more clear:
```java
Integer.toString(i)
```

nmuesch: Good call, that is definitely cleaner. Thanks!

PR #168, opened 2018-03-07
Add read timeout for how long we query beans

nmuesch:
Adds in the [`sun.rmi.transport.tcp.responseTimeout`](https://docs.oracle.com/javase/6/docs/technotes/guides/rmi/sunrmiproperties.html) system property before starting the RMI connection. 

Customer experienced a blocking bean that caused JMXFetch to hang indefinitely (order of magnitude hours) This should help reduce cases like that and allow JMXFetch to continue grabbing the remaining metrics.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/RemoteConnection.java
+++ b/src/main/java/org/datadog/jmxfetch/RemoteConnection.java
@@ -61,7 +61,7 @@ public class RemoteConnection extends Connection {
         }
         
         //Set an RMI timeout so we don't get stuck waiting for a bean to report a value
-        System.setProperty("sun.rmi.transport.tcp.responseTimeout", DEFAULT_RMI_RESPONSE_TIMEOUT);
+        System.setProperty("sun.rmi.transport.tcp.responseTimeout", rmi_timeout);
         
         createConnection();
 
--- a/src/main/java/org/datadog/jmxfetch/RemoteConnection.java
+++ b/src/main/java/org/datadog/jmxfetch/RemoteConnection.java
@@ -21,7 +21,7 @@ public class RemoteConnection extends Connection {
     private String rmi_timeout;
     private static final String TRUST_STORE_PATH_KEY = "trust_store_path";
     private static final String TRUST_STORE_PASSWORD_KEY = "trust_store_password";
-    private static final String DEFAULT_RMI_RESPONSE_TIMEOUT = "600000"; //Match the refresh_beans default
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/RemoteConnection.java
```diff
@@ -53,6 +54,10 @@ public RemoteConnection(LinkedHashMap<String, Object> connectionParams)
             }
 
         }
+        
+        //Set an RMI timeout so we don't get stuck waiting for a bean to report a value
+        System.setProperty("sun.rmi.transport.tcp.responseTimeout", DEFAULT_RMI_RESPONSE_TIMEOUT);
```
olivielpeau: what happens when we reach the timeout? Do we catch and handle the exception that's raised correctly?

src/main/java/org/datadog/jmxfetch/RemoteConnection.java
```diff
@@ -20,6 +20,7 @@
     private String jmx_url;
     private static final String TRUST_STORE_PATH_KEY = "trust_store_path";
     private static final String TRUST_STORE_PASSWORD_KEY = "trust_store_password";
+    private static final String DEFAULT_RMI_RESPONSE_TIMEOUT = "5000";
```
olivielpeau: based on the docs this is expressed in milliseconds, are you sure 5 seconds is enough in usual cases? (I know some JVMs expose a lot of beans, and on these environments I could imagine that the call to get all the beans takes more than 5 seconds).

src/main/java/org/datadog/jmxfetch/RemoteConnection.java
nmuesch: An error should be shown during the refresh bean list period where we scope out all available beans. If we hit the timeout there, an exception will be thrown (with a read timeout) and we should wait until the next collection period to try again.

src/main/java/org/datadog/jmxfetch/RemoteConnection.java
nmuesch: I bumped this up to 15 by default and am using the user defined "refresh_beans" parameter to address this and make it configurable. What do you think?

nmuesch: I bumped this again to actually be the default of the refresh_bean timeout as I think that makes the most sense here.

PR #171, opened 2018-04-16
Allow the troubleshooting commands to use the IPC endpoint

arbll:
This PR makes it possible to give configs to `list_*` commands using the HTTPS IPC endpoint.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -136,6 +136,11 @@ public class App {
 
         App app = new App(config);
 
+        // Get config from the ipc endpoint for "list_*" actions
+        if (!config.getAction().equals(AppConfig.ACTION_COLLECT)) {
+            app.getJSONConfigs();
+        }
+
         // Initiate JMX Connections, get attributes that match the yaml configuration
         app.init(false);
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -136,6 +136,11 @@ public static void main(String[] args) {
 
         App app = new App(config);
 
+        // Get config from the ipc endpoint for "list_*" actions
+        if (!config.getAction().equals(AppConfig.ACTION_COLLECT)) {
```
truthbk: Wondering if this might break stuff on agent5? Can you verify? Maybe we should add a flag to enable this behavior?

src/main/java/org/datadog/jmxfetch/HttpClient.java
```diff
@@ -107,7 +107,7 @@ public HttpResponse request(String method, String body, String path) {
             con.setRequestMethod(method.toUpperCase());
             con.setRequestProperty("Authorization", "Bearer "+ this.token);
             con.setRequestProperty("User-Agent", USER_AGENT);
-            if (method.toUpperCase() == "GET") {
+            if (method.toUpperCase().equals("GET")) {
```
truthbk: 🍪

src/main/java/org/datadog/jmxfetch/App.java
arbll: It should not since the A5 does not give the IPC config parameters . https://github.com/DataDog/jmxfetch/blob/1e163cf616758b2df52ac311d0183cf39c2c8c9c/src/main/java/org/datadog/jmxfetch/App.java#L538
^ `getJSONConfigs` is a no-op in A5

PR #174, opened 2018-06-25
fix jmx service checks

arbll:
## What does this PR do

JMXfetch used the name of the configuration instead of the check name in the configuration to send service checks. 
The recent refactor on datadog-agent that allowed having multiple configs for jmx by giving name different name exposed the issue and broke service checks for jmx. This fixes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,8 @@ jdk:
 - oraclejdk7
 - openjdk7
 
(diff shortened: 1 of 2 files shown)


arbll:
- Yes, since they both have their own way of fetching configs this isn't too surprising. Agent5 allows for multiple config with the same name since it uses an `Iterator`. Agent6 needs to to have different names for configuration since we map the JSON to a dictionary.
- I'll double check everything but this PR changes the check name at the instantiation of the check. It should replicate over everything using the check name.

PR #180, opened 2018-09-17
Convert to library

mar-kolya:
Allow JMX fetch to be used as a library.

At this point this is more of an RFC and not intended to be merged directly. Main changes include:
* Make JMXFetch releasable on maven. This PR currently releases into temporary repo and this would need to change.
* Refactor code so `System.exit()` is never called from main run function.
* Add support for loading configuration from Java resources.
* Add additional parameters and constructor method to config object to allow creating it from the user of this library.


src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -24,8 +24,10 @@ public static void setup(Level level, String logLocation) {
             Logger.getRootLogger().addAppender(fa);
             LOGGER.info("File Handler set");
         } else {
-            System.out.println("Log location is not set, will output log to stdout.");
-            ConsoleAppender consoleAppender = new ConsoleAppender(new PatternLayout(LOGGER_LAYOUT));
+            ConsoleAppender consoleAppender = new ConsoleAppender(new PatternLayout(LOGGER_LAYOUT), logLocation);
```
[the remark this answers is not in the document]
mar-kolya: Yeah, I should not have passed logLocation there, it is handled later. Is this what you meant the problem is?

src/main/java/org/datadog/jmxfetch/reporter/ConsoleReporter.java
```diff
@@ -61,24 +64,24 @@ public void doSendServiceCheck(String checkName, String status, String message,
...
     @Override
     public void displayMetricReached() {
-        System.out.println("\n\n\n       ------- METRIC LIMIT REACHED: ATTRIBUTES BELOW WON'T BE COLLECTED -------\n\n\n");
+        LOGGER.debug("       ------- METRIC LIMIT REACHED: ATTRIBUTES BELOW WON'T BE COLLECTED -------");
     }
 
     @Override
     public void displayMatchingAttributeName(JMXAttribute jmxAttribute, int rank, int limit) {
-        System.out.println("       Matching: " + rank + "/" + limit + ". " + jmxAttribute);
+        LOGGER.debug("       Matching: " + rank + "/" + limit + ". " + jmxAttribute);
```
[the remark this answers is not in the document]
mar-kolya: ConsoleReporter is accessible via tracing agent as well, so it would be nice to be consistent with it. I'll make this a warning.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -130,9 +140,41 @@ public Instance(LinkedHashMap<String, Object> instanceMap, LinkedHashMap<String,
             }
         }
 
-        // Add the configuration to get the default basic metrics from the JVM
-        configurationList.add(new Configuration((LinkedHashMap<String, Object>) new YamlParser(this.getClass().getResourceAsStream("/jmx-1.yaml")).getParsedYaml()));
-        configurationList.add(new Configuration((LinkedHashMap<String, Object>) new YamlParser(this.getClass().getResourceAsStream("/jmx-2.yaml")).getParsedYaml()));
+        loadMetricConfigFiles(appConfig, configurationList);
+
+        ArrayList<LinkedHashMap<String, Object>> defaultConf = (ArrayList<LinkedHashMap<String, Object>>) new Yaml().load(this.getClass().getResourceAsStream("default-jmx-metrics.yaml"));
```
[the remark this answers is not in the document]
mar-kolya: This change specifically just replaces jmx-1 and jmx-2 with single configuration file, and extends that configuration slightly. So this is sort of preexisting behavior.
Do you see a problem with this?

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -495,40 +500,68 @@ public boolean addJsonConfig(String name, String json) {
...
-        YamlParser fileConfig;
 
+        loadFileConfigs(config, configs);
+        loadResourceConfigs(config, configs);
+
+        LOGGER.info("Found " + configs.size() + " config files");
+        return configs;
+    }
+
+    private void loadFileConfigs(AppConfig config, ConcurrentHashMap<String, YamlParser> configs) {
```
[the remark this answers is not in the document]
mar-kolya: Well, 'load files' was a preexisting functionality - I've just refactored some code into this helper function.
The new thing is below - loading files via resources. And this is kind of useful because this allows APM agent to package resource with itself and pass it to jmxfetch without too much hassle.

src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
```diff
@@ -0,0 +1,145 @@
...
+        metric_type: gauge
+      SystemCpuLoad:
+        alias: jvm.cpu_load.system
+        metric_type: gauge
+
+# Young Gen Collectors (Minor Collections)
+- include:
+    domain: java.lang
+    type: GarbageCollector
+    name: Copy
```
arbll: Not seeing a mention to this filter in https://docs.datadoghq.com/integrations/java/#description-of-the-filters. What does it do ? And how do we differenciate `jvm.gc.minor_collection_count` from `Copy` vs `ParNew` for ex ?

src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
```diff
@@ -0,0 +1,145 @@
+# Memory
```
arbll: I see multiple differences between `default-jmx-metrics.yaml` and the default metrics that we used to collect in `jmx-1` and `jmx-2` . It's very important that we don't remove/rename any metrics as it would break the graphs, monitors, etc built around them. 

Here are the diffs that I found Before => Now :
- `jvm.thread_count` => `thread.count`
- `jvm.gc.parnew.time` => `jvm.gc.minor_collection_time` ?
- `jvm.gc.cms.count` => `jvm.gc.minor_collection_count` ?

There are also multiple new metrics introduced, do we need to add them and would it be possible to move them to another PR ? That way we can focus on keeping compatibility. Improving the testing here might make sense.

src/test/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
```diff
@@ -0,0 +1,43 @@
+# This is a reduced set of 'default' metrics to make tests more predictable
```
arbll: Same as above, let's keep compatibility with the metric names. You can use https://docs.datadoghq.com/integrations/java/#metrics

pom.xml
```diff
@@ -1,71 +1,86 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
 
-	<groupId>datadog</groupId>
-	<artifactId>jmxfetch</artifactId>
-	<version>0.20.2</version>
-	<packaging>jar</packaging>
+    <groupId>datadog</groupId>
+    <artifactId>jmxfetch</artifactId>
+    <version>0.20.3</version>
```
arbll: We'll bump that in another PR (and probably increase the minor since it's a pretty big change 😛 )

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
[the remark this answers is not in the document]
arbll: If you revert this line, this looks good to me.

src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
mar-kolya: Thad documentation says `On top of these parameters, the filters support “custom” keys which means that you can filter by bean parameters. ` - which is what we are using here. We are just filtering by specific GC name.

arbll: Okay I understand now, seems like I missed that part 😅 . Now for the second part of my comment, I guess you can only have one `Young Gen Collector` and one `Old Gen Collector` ?

mar-kolya: Yes, JVM allows only one Old and one New gen collector - by there are a few different possibly types of those.

src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
mar-kolya: I've pushed that change - please let me know if this is an acceptable solution.

src/test/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
mar-kolya: This file is actually use only by tests, so it is not terribly important what specific metrics are inhere.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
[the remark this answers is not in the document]
mar-kolya: Yes, done. Sorry, I've reverted that a while back but apparently pushed into a wrong place. My bad, fixed now.

src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
arbll: Thanks for that. This sounds acceptable and adding 4 new metrics should not be an issue. This would also have the nice side effect of allowing us to remove some complexity from the config files of other jmx integrations like cassandra.
We should remove `jvm.gc.parnew.time` and `jvm.gc.cms.count` from the docs and add the new ones after this is merged.

src/test/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
arbll: Yes I know but I was hoping that the tests would catch the metric name changes to avoid regressions. Did they not ?

src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
arbll: I'll discuss this changes with my team to see if there is anything I am missing on why this was not added previously. I'll then do some sanity check with A5 and A6 and if everything looks good I'll merge

src/test/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
mar-kolya: I had to make some changes to tests to make build pass... So I guess they did to some extent.

pom.xml
mar-kolya: Should be fixed now.

PR #184, opened 2018-10-01
Support the empty_default_hostname instance field

xvello:
This makes jmxfetch submit metrics and service-checks with the dummy `host:` tag for cluster checks. It is supported for metrics since agent5, and for service-checks in 6.6.

See https://github.com/DataDog/datadog-agent/pull/2334 and https://github.com/DataDog/datadog-agent/pull/2351


src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -51,7 +51,8 @@
     private Boolean cassandraAliasing;
 
     JMXAttribute(MBeanAttributeInfo attribute, ObjectName beanName, String instanceName,
-            Connection connection, HashMap<String, String> instanceTags, Boolean cassandraAliasing) {
+            Connection connection, HashMap<String, String> instanceTags, Boolean cassandraAliasing,
```
truthbk: How do you feel we should handle the case where the instanceTags `host` tag is set (for whatever reason), but `emptyDefaultHostname` is true. IMHO it might make sense to just evict it from the HashMap and make the boolean option take precedence.

src/test/java/org/datadog/jmxfetch/TestInstance.java
```diff
@@ -1,34 +1,75 @@
...
+		registerMBean(new SimpleTestJavaApp(), "org.datadog.jmxfetch.test:foo=Bar,qux=Baz");
+		initApplication("jmx_empty_default_hostname.yaml");
+		run();
+
+		LinkedList<HashMap<String, Object>> metrics = getMetrics();
+		assertEquals(28, metrics.size());
+		for (HashMap<String, Object> metric : metrics) {
+			String[] tags = (String[]) metric.get("tags");
+			this.assertHostnameTags(Arrays.asList(tags));
+    	}
```
truthbk: nit: this doesn't look like it's properly aligned 📏

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -83,10 +84,16 @@ public Instance(LinkedHashMap<String, Object> instanceMap, LinkedHashMap<String,
             // a jmxtree that is not completely initialized and would be missing some attributes
         }
 
-        this.minCollectionPeriod = (Integer) instanceMap.get("min_collection_interval");
+        this.minCollectionPeriod = (Integer) this.instanceMap.get("min_collection_interval");
```
truthbk: Not really your fault, but we have so much room for `NullPointerException` in this constructor.....

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -83,10 +84,16 @@ public Instance(LinkedHashMap<String, Object> instanceMap, LinkedHashMap<String,
...
         if (this.minCollectionPeriod == null && initConfig != null) {
         	this.minCollectionPeriod = (Integer) initConfig.get("min_collection_interval");
         }
+
+        this.emptyDefaultHostname = (Boolean) this.instanceMap.get("empty_default_hostname");
```
truthbk: Again, not your fault as you're following a pattern used in the class, but I'd say it's safer to use `boolean` as the type for `this.emptyDefaultHostname`.

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
xvello: For python/golang checks, we want to support an empty default hostname + check-provided hostname override (for kubernetes events node events / kube_state node metrics), so it's a legit use case.

It's currently not possible as an instance tag as [it'll be sanitized to bean_host](https://github.com/DataDog/jmxfetch/blob/master/src/main/java/org/datadog/jmxfetch/JMXAttribute.java#L165), but could it be added as a bean tag? 

If so, do you think there's a better method to plug this logic into?

src/main/java/org/datadog/jmxfetch/Instance.java
xvello: Indeed, though the same thing while working on it. Do you think we should wrap these in try/catch blocks?

xvello: What about:
```
    private boolean emptyDefaultHostname;

...

        try {
            this.emptyDefaultHostname = (Boolean) this.instanceMap.get("empty_default_hostname");
        } catch (NullPointerException e) {
            this.emptyDefaultHostname = false;
        }
```

xvello:
@truthbk rebased and addressed your feedback

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
truthbk: Let's leave as-is, I think its probably fine.

src/test/java/org/datadog/jmxfetch/TestInstance.java
```diff
@@ -1,34 +1,75 @@
...
+		} else if (tagList.contains(new String("instance:jmx_test_no_hostname"))) {
+			// empty_default_hostname case
+			assertTrue(tagList.contains(new String("host:")));
+		} else {
+			fail("unexpected instance tag");
+		}
+	}
+
+	@Test
+	public void testEmptyDefaultHostname() throws Exception {
```
truthbk: 🍰

PR #185, opened 2018-10-03
add support for rmi registry connection over ssl

arbll:
### Purpose of this PR

We were previously only supporting SSL for connections to RMI servers and not for connections to the RMI registry. 

Users will most likely want to authenticate their JMX clients using 2-way ssl as this is recommended here https://docs.oracle.com/javase/8/docs/technotes/guides/management/agent.html. 
This means we also need to use a keystore for the client, which is why I added two new config options.

### Motivation

Client request


olivielpeau:
very nice 👌 

I might be stating the obvious, but let's document these new options in the `jmx.yaml.example` of both Agent 5 & 6 and update the docs website

PR #190, opened 2018-10-12
Add support for metric_type gauge_delta

mirkoprescha:
This PR allows to define "gauge_delta" as new metric_type in jmx.d/conf.yaml. JMXFetch will send the metric as StatsD Count-Type which will be handled as "rate" in DataDog. This allows to do time aggregations like "how many errors we had in last hour".

It might be also possible to use/reuse metric_type counter / count, but I am not sure about breaking semantics. Actually this PR is a POC of how gauge deltas can be implemented, as we really need them.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -413,6 +413,7 @@ public class App {
 
             // Clearing rates aggregator so we won't compute wrong rates if we can reconnect
             reporter.clearRatesAggregator(instance.getName());
+            reporter.clearCountersAggregator(instance.getName());
 
             LOGGER.warn("Instance " + instance + " didn't return any metrics." +
                     "Maybe the server got disconnected ? Trying to reconnect.");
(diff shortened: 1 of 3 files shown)


arbll:
Hi @mirkoprescha!

Thanks a lot for this proof of concept. A way to send counts to datadog is clearly missing in jmxfetch. 

I had some testing to do to validate the approach so I took the liberty of extending this PR at https://github.com/DataDog/jmxfetch/tree/arbll/count-support. I mostly added tests and renamed `gauge_delta` to `monotonic_count` since what we do here is very similar to the `monotonic_count` [from the agent checks](https://docs.datadoghq.com/developers/metrics/counts/#agent-check). 
I also removed the use of the `canonicalRate` flag and kept the behavior when it was equal to `true` as it's something we would like to enable by default for rates but can't since it would break compatibility. As a side note, this means that there won't be any way to report negative deltas.

Apart from these small changes, everything looks good! I'll likely open another PR based on the branch linked above. If nothing comes up it should be merged soon and will be included the version 6.9 of the agent.

Best,
Arthur

truthbk:
As mentioned by @arbll, this we've based #203 off of this great work by @mirkoprescha, so we will close this PR and piggy back off of it in the mentioned PR. Please follow discussion there. Thank you so much for your contribution Mirko! 🙇

PR #191, opened 2018-10-18
allow setting extra tags when using as library

mar-kolya:


mar-kolya:
Rebased

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -181,21 +181,22 @@ private void loadMetricConfigFiles(AppConfig appConfig, LinkedList<Configuration
...
         }
-        catch (ClassCastException e){
-            // Input has `List` format
-            LinkedHashMap<String, String> tags = new LinkedHashMap<String, String>();
-
-            for (String tag: (List<String>)tagsMap) {
-                tags.put(tag, null);
+        if (tagsMap != null) {
+            try {
+                // Input has `Map` format
```
arbll: Nit: I guess you could have left `// Input has `List` format`.

mar-kolya: Updated

PR #197, opened 2018-11-22
put new gc metrics behind a flag

arbll:
In 6.6 we introduced new names for gc metrics :
* `jvm.gc.minor_collection_count`
* `jvm.gc.minor_collection_time`
* `jvm.gc.major_collection_count`
* `jvm.gc.major_collection_time`

 This names make more sense but unfortunately due to some gotchas in the way jmxfetch collect metrics this also had the effect of not sending the metrics using the old names.

This PR fixes this by putting the new names behind a flag.


olivielpeau:
(tracking this flag as something we could want to enable by default in version 7 of the Agent)

olivielpeau:
Had a couple of questions/notes:

* how would this PR affect JMXFetch in the APM java agent?
* (not urgent): we should document this new option in the `jmx.d/conf.yaml.example` file that's shipped with the agent, and also in all the example conf files of integrations

truthbk:
Synced with APM, they think it's fine, but they will schedule some work on their side to make sure they have full compatibility with either collection mode.

PR #198, opened 2018-11-26
Default JVM metrics for GC pools, class load count, and descriptors

realark:
For our APM JMX graphs we're interested in adding charts for specific GC pools (eden, old gen, etc), class load counts, and open file descriptors.

Couple of questions:

* Does it make sense to put these changes here (in core JMX)?
* Process-wise, does it work to propose built-in metric changes like this (through a PR)?

### DIFF
--- a/src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
+++ b/src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
@@ -49,3 +49,94 @@
         alias: jvm.cpu_load.system
         metric_type: gauge
 
+# Classloading
+- include:
+    domain: java.lang
+    type: ClassLoading
+    attribute:
+      LoadedClassCount:
+        alias: jvm.loaded_classes
+        metric_type: gauge
+
+# Open File Descriptors
+- include:
+    domain: java.lang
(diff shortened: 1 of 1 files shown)


src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
```diff
@@ -49,3 +49,94 @@
         alias: jvm.cpu_load.system
         metric_type: gauge
 
+# Classloading
+- include:
+domain: java.lang
```
olivielpeau: wrong indentation for this block

realark: Thanks! Must have happened when I resolved a conflict. Fixed now.

realark:
Excellent, thanks! Yes it was my intent to add these to the default metrics.

These metrics will be used by APM to generate interesting JVM graphs and it seems like they are interesting to agent jmxfetch users as well.

olivielpeau:
@realark we can merge this if you've confirmed it works well :)

realark:
@olivielpeau Thanks! Yep it's working well! Just did some more end-to-end testing to verify:

image

PR #204, opened 2018-12-27
Replace $value by the JMX value

danosipov:
This PR enables the use of metric value as part of the metric name.

This would probably be most useful for string value metrics together with value conversion so that the double value is submitted correctly to DD.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
@@ -506,7 +506,6 @@ public abstract class JMXAttribute {
             // Same as above
         }
 
-
         return alias;
     }
 
--- a/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
@@ -503,6 +503,12 @@ public abstract class JMXAttribute {
         // Attribute & domain
         alias = alias.replace("$attribute", fullAttributeName);
         alias = alias.replace("$domain", domain);
+        try {
+            alias = alias.replace("$value", getJmxValue().toString());
+        } catch (Exception e) {
+            // If we haven't been able to get the value, it wasn't replaced.
+        }
+
 
         return alias;
     }
--- a/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
+++ b/src/main/java/org/datadog/jmxfetch/JMXAttribute.java
@@ -15,12 +15,7 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import javax.management.AttributeNotFoundException;
-import javax.management.InstanceNotFoundException;
-import javax.management.MBeanAttributeInfo;
-import javax.management.MBeanException;
-import javax.management.ObjectName;
-import javax.management.ReflectionException;
+import javax.management.*;
 
 import org.apache.log4j.Logger;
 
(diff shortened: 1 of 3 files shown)


danosipov:
@olivielpeau Addressed.

I think this could be potentially misused to cause metric spam, but that's a general danger that's not specific to this change.

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -503,6 +498,13 @@ private String getUserAlias(LinkedHashMap<String, LinkedHashMap<String, String>>
         // Attribute & domain
         alias = alias.replace("$attribute", fullAttributeName);
         alias = alias.replace("$domain", domain);
+        try {
+            alias = alias.replace("$value", getJmxValue().toString());
+        } catch (JMException e) {
+            // If we haven't been able to get the value, it wasn't replaced.
```
olivielpeau: we should log a warning here

src/main/java/org/datadog/jmxfetch/JMXAttribute.java
```diff
@@ -503,6 +498,13 @@ private String getUserAlias(LinkedHashMap<String, LinkedHashMap<String, String>>
...
+        } catch (IOException e) {
+            // Same as above
```
olivielpeau: ditto

src/test/java/org/datadog/jmxfetch/TestApp.java
```diff
@@ -100,6 +100,7 @@ public void testRegexpAliasing() throws Exception {
 
         assertMetric("this.is.100.bar.baz", tags, 4);
         assertMetric("org.datadog.jmxfetch.test.baz.hashmap.thisis0", tags, 4);
+        assertMetric("this.is.thousand.1000.0", 1000, tags, 4);
```
olivielpeau: what's the actual value of the metric when the JMX attribute's value is a string? Could you add a test on that? (you can add attributes to the test app that exposes JMX attributes)

AFAIK this feature is only really useful if the actual value is a string actually

danosipov:
@olivielpeau Addressed, let me know what you think

src/test/java/org/datadog/jmxfetch/TestApp.java
```diff
@@ -733,8 +735,7 @@ public void testServiceDiscovery() throws Exception {
         LinkedList<HashMap<String, Object>> metrics = getMetrics();
 
         // 14 = 13 metrics from java.lang + 1 metric explicitly defined in the yaml config file
-        assertEquals(59, metrics.size());
-
+        assertEquals(63, metrics.size());
```
olivielpeau: do you know why the number of metrics increased by 4 here? (haven't checked in details, but my expectation would be an increase of 2)

(same question on the other test below)

danosipov: Looks like each of the two metrics is submitted with different tags:
```
[type:SimpleTestJavaApp, component, scope:CoolScope, instance:jmx_test_instance, jmx_domain:org.datadog.jmxfetch.test, bean_host:localhost]
[foo:Bar, qux:Baz, instance:jmx_test_instance, jmx_domain:org.datadog.jmxfetch.test]
```

I'm guessing its because of how its registered here:
https://github.com/DataDog/jmxfetch/pull/204/files#diff-974534e55eb97868b68abda77ce93ecbR728
but I'm not too familiar with the tests

olivielpeau: makes sense, thanks for checking 👍

PR #205, opened 2019-01-02
 Allow metrics configs to be loaded via resources

tylerbenson:
Metrics configs embedded inside the agent aren’t directly accessible on the file system.  Loading as a resource avoids having to copy to an external location.

These files should match the standard `metrics.yaml` format, and include `jmx_metrics` blocks at the top level.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -6,7 +6,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Enumeration;
@@ -27,7 +27,6 @@ import javax.security.auth.login.FailedLoginException;
 import org.apache.log4j.Appender;
 import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
-import org.apache.commons.lang3.CharEncoding;
 import org.apache.commons.io.IOUtils;
 import org.datadog.jmxfetch.reporter.Reporter;
 import org.datadog.jmxfetch.util.CustomLogger;
@@ -51,6 +50,7 @@ public class App {
     private static final String AD_LEGACY_CONFIG_TERM = "#### SERVICE-DISCOVERY TERM ####";
     private static final int AD_MAX_NAME_LEN = 80;
     private static final int AD_MAX_MAG_INSTANCES = 4; // 1000 instances ought to be enough for anyone
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
 
     private static int loopCounter;
     private int lastJSONConfigTS;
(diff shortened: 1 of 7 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -220,7 +219,7 @@ public boolean processAutoDiscovery(byte[] buffer) {
       String[] discovered;
 
       try {
-        String configs = new String(buffer, CharEncoding.UTF_8);
+        String configs = new String(buffer, "UTF-8");
```
gzussa: Good catch to remove this dependency since it was only used for this. 
What about creating constant instead of having duplicated strings?

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -38,6 +39,13 @@
     public static final String PROCESS_NAME_REGEX = "process_name_regex";
     public static final String ATTRIBUTE = "Attribute: ";
 
+    private static final ThreadLocal<Yaml> YAML = new ThreadLocal<Yaml>() {
```
gzussa: Why does it need to be a `ThreadLocal` object?

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -199,6 +208,40 @@ private void loadMetricConfigFiles(AppConfig appConfig, LinkedList<Configuration
         }
     }
 
+    private void loadMetricConfigResources(AppConfig config, LinkedList<Configuration> configurationList) {
```
gzussa: Is it legit to add a unit tests for this method, unless it is covered already?

src/main/java/org/datadog/jmxfetch/App.java
tylerbenson: I can do that if that's what you'd prefer.

src/main/java/org/datadog/jmxfetch/Instance.java
tylerbenson: There seemed to be non-trivial amount of initialization happening in the constructor, so I wanted to avoid it.  The problem is (I believe) `Yaml` is not thread safe, so I used the ThreadLocal to ensure that safety.

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -51,6 +50,7 @@
     private static final String AD_LEGACY_CONFIG_TERM = "#### SERVICE-DISCOVERY TERM ####";
     private static final int AD_MAX_NAME_LEN = 80;
     private static final int AD_MAX_MAG_INSTANCES = 4; // 1000 instances ought to be enough for anyone
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
```
gzussa: Nice refactoring! 👍

src/main/java/org/datadog/jmxfetch/Instance.java
gzussa: ok I trust you judgement here

src/main/java/org/datadog/jmxfetch/Instance.java
gzussa: You most probably will have to create a method to get the `inputStream` so this method can be testable as such by creating an object instance and calling this method while providing your inputStream.

tylerbenson: Ok, I've added a few simple tests. Is this what you were looking for?

tylerbenson:
@gzussa would you like to merge this or should I request access to merge it myself?

PR #207, opened 2019-01-23
Upgrade java-dogstatsd-client to 2.7

mattdrees:
This PR upgrades java-dogstatsd-client from 2.1.0 to 2.7.

We'd like to take advantage of https://github.com/DataDog/java-dogstatsd-client/pull/42, in particular.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,7 @@
         <commons-lang.version>2.6</commons-lang.version>
         <apache-commons-lang3.version>3.5</apache-commons-lang3.version>
         <guava.version>17.0</guava.version>
-        <java-dogstatsd-client.version>2.1.0</java-dogstatsd-client.version>
+        <java-dogstatsd-client.version>2.7</java-dogstatsd-client.version>
         <jcommander.version>1.35</jcommander.version>
         <log4j.version>1.2.17</log4j.version>
         <jackson.version>2.9.8</jackson.version>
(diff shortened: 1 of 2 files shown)


mattdrees:
Rebased to resolve conflicts and match new style guide.

mattdrees:
@truthbk could I bother you to review this PR? It's not urgent, but I'm hoping it can be included in your next release, whenever that may be.

mattdrees:
This is our use case:
(copy/pasting from my support ticket response, because it might be helpful here)

We run a java application in a docker container orchestrated by AWS ECS. We'd like to collect metrics from it, as well as APM traces.

Your docs recommend using unix domain sockets for collecting metrics from containerized apps: https://docs.datadoghq.com/developers/dogstatsd/unix_socket/
Sounds like a good idea to us.

Your dd-trace-java library is needed for APM, and it also embeds JMXFetch so we can use it for metrics, without having to run it separately within the container (which would be a pain).
So using the embedded JMXFetch tool sounds great.

Unfortunately, JMXFetch doesn't support unix domain sockets, because it uses an older java-dogstatsd version. So we can't currently follow your recommendations. (Except by setting up socat, which works but requires another process to run inside my container.)

mattdrees:
@olivielpeau could I ask you to review?

mattdrees:
Thanks Arthur!

PR #210, opened 2019-02-25
Small logging + logic clean up of concurrent collection

olivielpeau:
### What this PR does

See individual commit descriptions. Overall:

* fbf7423: improvements to log entries
* 88ed5d4: remove unused logic on `CachedThreadPool` creation, since the config code doesn't allow negative values for the number of threads at the moment. A future PR (for `0.27.0`) will allow negative values.
* 2180909: small fix to TaskProcessor readiness logic: before the first JMXFetch "iteration", `tpe.getPoolSize()` is `0`, so the app would consider the pool as not ready and print out warnings + recreate the thread pool as long the first collection isn't actually kicked off.

### Notes

`App.processFixedStatus` doesn't log anything at the moment, not sure whether logging from there would potentially spam a lot of logs, would appreciate suggestions on that.

Overall, 2180909 requires particular attention as it changes actual logic, but the other changes shouldn't require any QA.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -450,7 +450,9 @@ public class App {
             }
 
             if (!collectionProcessor.ready()) {
-                LOGGER.warn("Executor has to be replaced, previous one hogging threads");
+                LOGGER.warn(
+                        "Executor has to be replaced for collection processor, "
+                        + "previous one hogging threads");
                 collectionProcessor.stop();
                 collectionProcessor.setThreadPoolExecutor(
                         Executors.newFixedThreadPool(appConfig.getThreadPoolSize()));
@@ -786,7 +788,6 @@ public class App {
         Iterator<Entry<String, YamlParser>> it = configs.entrySet().iterator();
         Iterator<Entry<String, YamlParser>> itPipeConfigs = adPipeConfigs.entrySet().iterator();
         while (it.hasNext() || itPipeConfigs.hasNext()) {
-            LOGGER.info("iterating statics...");
             Map.Entry<String, YamlParser> entry;
             boolean fromPipeIterator = false;
             if (it.hasNext()) {
@@ -884,7 +885,7 @@ public class App {
             processStatus(instanceInitTasks, statuses);
         } catch (Exception e) {
             // NADA
-            LOGGER.warn("critical issue initializing instances: " + e);
+            LOGGER.warn("Critical issue initializing instances: " + e);
         }
     }
 
@@ -973,9 +974,11 @@ public class App {
 
                 // All was good, add instance
                 instances.add(instance);
-                LOGGER.info("Recovered broken instance: " + idx);
+                LOGGER.info("Successfully initialized instance: " + instance.getName());
             } catch (Throwable e) {
-                LOGGER.info("Instance remains broken: " + instance.getName());
+                LOGGER.info(
+                    "Could not initialize instance: " + instance.getName()
+                    + ": " + e.toString());
                 instance.cleanUpAsync();
                 brokenInstanceMap.put(instance.toString(), instance);
             }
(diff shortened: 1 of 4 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -85,16 +85,12 @@ public App(AppConfig appConfig) {
         this.appConfig = appConfig;
 
         ExecutorService collectionThreadPool =
-                appConfig.getThreadPoolSize() >= 1
-                        ? Executors.newFixedThreadPool(appConfig.getThreadPoolSize())
-                        : Executors.newCachedThreadPool();
+                Executors.newFixedThreadPool(appConfig.getThreadPoolSize());
```
truthbk: So one of the reasons we have this is because if customers set a negative value for the thread pool size this would be the same as having an unlimited pool. I guess you want to avoid that? It is dangerous, but could be useful to debug.

src/main/java/org/datadog/jmxfetch/tasks/TaskProcessor.java
```diff
@@ -38,7 +38,7 @@ public void setThreadPoolExecutor(ExecutorService executor) {
      * */
     public boolean ready() {
         ThreadPoolExecutor tpe = (ThreadPoolExecutor) threadPoolExecutor;
-        return !(tpe.getPoolSize() == tpe.getActiveCount());
+        return !(tpe.getMaximumPoolSize() == tpe.getActiveCount());
```
truthbk: I think for a fixed thread pool these two are equivalent, but this is better and less error-prone as we also had some cached thread-pool support (although you've killed it).

src/test/java/org/datadog/jmxfetch/tasks/TestTaskProcessor.java
```diff
@@ -59,7 +59,7 @@ static TaskStatusHandler processTestResults(
             }
         } catch (Exception e){
             exc = e;
-        } 
+        }
```
truthbk: 🍰

src/main/java/org/datadog/jmxfetch/tasks/TaskProcessor.java
olivielpeau: should've mentioned this earlier, but for a fixed thread pool this isn't equivalent: before the first JMXFetch "iteration", `tpe.getPoolSize()` is `0`, so the app considers the pool as not ready and prints out warnings before the first collection is actually kicked off.

src/main/java/org/datadog/jmxfetch/App.java
olivielpeau: as I mentioned in the commit description: `we validate that the configured number of threads is strictly positive`. So we don't allow users to set a `0` or negative value here

src/main/java/org/datadog/jmxfetch/tasks/TaskProcessor.java
truthbk: Interesting. Nice catch.

src/main/java/org/datadog/jmxfetch/App.java
truthbk: 🤦‍♂️ OK then let's decide if we want to kill the feature or fix the config value.

truthbk: decision: we might implement an unbounded pool feature in the future, no need to do it now.

olivielpeau: discussed on slack: let's plan allowing negative values for jmxfetch `0.27.0`, only allowing strictly positive values is enough for `0.26.1` (users can set a very high value on the number of threads, and it'll be very close to using an unbounded thread pool)

olivielpeau:
Updated PR description with our inline discussion.

Merging!

PR #223, opened 2019-05-08
Log via slf4j instead of directly to log4j

tylerbenson:
The logs are still directed to log4j, but this allows dd-java-agent to exclude log4j and send logs through its’ log system.

Here I’m using the lombok annotation as an optimization, but I configured it to use the same field name for consistency (and reduce the delta).


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -213,27 +221,16 @@ public static int run(AppConfig config) {
...
-                                        }
-                                    }
-                                });
+        Runtime.getRuntime().addShutdownHook(
+            new Thread() {
+                @Override
+                public void run() {
+                    LOGGER.info("JMXFetch is closing");
+                    // Properly close log handlers
+                    LogManager.shutdown();
```
tylerbenson: This was unnecessarily complicated (unless I'm missing something).  Hope you don't mind the cleanup.

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -1104,7 +1101,7 @@ static TaskStatusHandler processCollectionResults(
                                     + " metrics.";
 
                     instanceStatus = Status.STATUS_WARNING;
-                    CustomLogger.laconic(LOGGER, Level.WARN, instanceMessage, 0);
+                    CustomLogger.laconic(LOGGER, instanceMessage, 0);
```
tylerbenson: Slf4j doesn't support passing in a log level as an argument, but warn was the only thing being passed in, so I just made that the default value.

src/main/java/org/datadog/jmxfetch/App.java
truthbk: Definitely not! JMXFetch is known to have quite a bit of code that could use some cleanup, so this is definitely welcome.

src/main/java/lombok.config
```diff
@@ -0,0 +1 @@
+lombok.log.fieldName=LOGGER
```
truthbk: Could we add a single-line comment here explaining how this play with and Lombok and how this will generate a `LOGGER` attribute instead of the default `log` one?

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -41,9 +42,9 @@ public static void setup(Level level, String logLocation) {
     }
 
     /** Laconic logging for reduced verbosity. */
-    public static void laconic(Logger logger, Level level, String message, int max) {
+    public static void laconic(org.slf4j.Logger logger, String message, int max) {
```
truthbk: I'm pretty sure you're going to hate this suggestion, but I'm going to throw this out there just in case. What do you think about using reflection to call the right method depending on `Level level`? That way we could keep log level support for the laconic logging.

pom.xml
```diff
@@ -120,6 +120,22 @@
             <artifactId>jackson-core</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
```
truthbk: Wondering if we should document somewhere how Lombok annotations will "generate" some code for us. Mostly to help contributors not familiar with the library understand where LOGGER is getting instantiated. Perhaps a small note in the README? Maybe overkill if we don't plan to introduce lombok further into the codebase.

src/main/java/lombok.config
tylerbenson: I would prefer to remove it and replace all the calls with log instead if that’s ok.

pom.xml
tylerbenson: done.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
tylerbenson: I decided to use a switch statement instead.

PR #224, opened 2019-05-15
fix for eclipse

jdyer1:
When importing this repo as an "existing maven project" into eclipse, the source checkout is dirty.  The .settings directory is not in .gitignore, and someone accidently committed their .settings folder in the past.

- remove eclipse .settings folder from source control
- add .settings/ to .gitignore


albertvaka:
Thanks for noticing this!

PR #225, opened 2019-05-16
Use Lombok Builder for AppConfig, Formalize jvm_direct Connection

tylerbenson:
Based on #223, so focus on last two commits for review.

Will be used in `dd-java-agent`. A future release can completely remove the `metricConfigFiles` setting (but not `metricConfigResources`).

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -830,7 +830,7 @@ public class App {
 
             for (LinkedHashMap<String, Object> configInstance : configInstances) {
                 if (appConfig.isTargetDirectInstances() != isDirectInstance(configInstance)) {
-                    log.debug("Skipping instance '{}'. targetDirectInstances={} jvm_direct={}",
+                    log.info("Skipping instance '{}'. targetDirectInstances={} != jvm_direct={}",
                             name,
                             appConfig.isTargetDirectInstances(),
                             isDirectInstance(configInstance));
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -829,10 +829,10 @@ public class App {
             }
 
             for (LinkedHashMap<String, Object> configInstance : configInstances) {
-                if (appConfig.isAllowDirectInstances() != isDirectInstance(configInstance)) {
-                    log.debug("Skipping instance '{}'. allowDirectInstances={} jvm_direct={}",
+                if (appConfig.isTargetDirectInstances() != isDirectInstance(configInstance)) {
+                    log.debug("Skipping instance '{}'. targetDirectInstances={} jvm_direct={}",
                             name,
-                            appConfig.isAllowDirectInstances(),
+                            appConfig.isTargetDirectInstances(),
                             isDirectInstance(configInstance));
                     continue;
                 }
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -110,7 +110,7 @@ public class App {
     public static void main(String[] args) {
 
         // Load the config from the args
-        AppConfig config = new AppConfig();
+        AppConfig config = AppConfig.builder().build();
         JCommander commander = null;
         try {
             // Try to parse the args using JCommander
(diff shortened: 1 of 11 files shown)


src/main/java/org/datadog/jmxfetch/AppConfig.java
```diff
@@ -204,6 +204,13 @@
     @Builder.Default
     private int ipcPort = 0;
 
+    /**
+     * Boolean setting to determine whether to ignore jvm_direct instances.
+     * If set to true, other instances will be ignored.
+     */
+    @Builder.Default
+    private boolean allowDirectInstances = false;
```
albertvaka: The "allow" in the name doesn't convey that `If set to true, other instances will be ignored.`.

Maybe something like `targetDirectInstances`?

tylerbenson: Yeah, I was having a hard time coming up with a better name... I like that better though.  Thanks for the suggestion.

tylerbenson: Done.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -41,9 +42,30 @@ public static void setup(Level level, String logLocation) {
     }
 
     /** Laconic logging for reduced verbosity. */
-    public static void laconic(Logger logger, Level level, String message, int max) {
+    public static void laconic(org.slf4j.Logger logger, Level level, String message, int max) {
```
truthbk: Yes! Glad you did this! 🍰

src/test/resources/jmx.yaml
```diff
@@ -1,7 +1,7 @@
 init_config:
 
 instances:
-    -   process_name_regex: .*surefire.*
+    -   jvm_direct: true
```
truthbk: Aren't `process_name_regex` and `jvm_direct` slightly different? One will attach to a running JVM, while the other is launched as a `-javaagent` with the application? Would different test cases be needed?

src/test/java/org/datadog/jmxfetch/tasks/TestTaskProcessor.java
```diff
@@ -1,37 +1,30 @@
...
 import org.junit.Test;
 import org.junit.BeforeClass;
 
 public class TestTaskProcessor {
     private static List<Instance> instances = null;
 
-	private final static Logger LOGGER = Logger.getLogger("Test Task Processor");
-
-    class TestSimpleTask extends InstanceTask<Boolean> {
+	class TestSimpleTask extends InstanceTask<Boolean> {
```
truthbk: nit: this doesn't look properly aligned?

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -824,14 +823,22 @@ public void init(boolean forceNewConnection) {
...
                 String warning = "No instance found in :" + name;
-                LOGGER.warn(warning);
+                log.warn(warning);
                 appConfig.getStatus().addInitFailedCheck(name, warning, Status.STATUS_ERROR);
                 continue;
             }
 
             for (LinkedHashMap<String, Object> configInstance : configInstances) {
+                if (appConfig.isTargetDirectInstances() != isDirectInstance(configInstance)) {
+                    log.debug("Skipping instance '{}'. targetDirectInstances={} jvm_direct={}",
```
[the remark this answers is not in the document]
tylerbenson: I'll change it to info.

src/test/resources/jmx.yaml
tylerbenson: They are slightly different, but from this test perspective, they're mostly the same.  There are plenty of other tests that verify the `process_name_regex` behavior, so I felt it was ok to change this one rather than add additional complexity in the tests.

(If we moved the all to `jvm_direct` we might be able to run the tests from within an IDE, rather than just via maven.)

src/test/java/org/datadog/jmxfetch/tasks/TestTaskProcessor.java
tylerbenson: turns out there was a sneaky tab in there...  🤷‍♂

PR #229, opened 2019-05-24
only log check names instead of their config

arbll:
### What does this PR do

Log updated list of checkName (it's actually `<checkName>_<configHash>`) instead of the complete JSON configs. The check name should be enough to link it to a template and get all the relevant informations.

Note that we were previously logging even if there wasn't an update. This seems useless.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -732,7 +732,7 @@ public class App {
                 update = true;
                 log.info("update is in order - updating timestamp: " + lastJsonConfigTs);
                 for (String checkName : adJsonConfigs.keySet()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -733,6 +731,9 @@ private boolean getJsonConfigs() {
                 lastJsonConfigTs = timestamp;
                 update = true;
                 log.info("update is in order - updating timestamp: " + lastJsonConfigTs);
+                for (String checkName : adJsonConfigs.keySet()) {
+                    log.debug("check " + checkName + " will be run");
```
olivielpeau: just to make the message a bit more explicit, what do you think of:
```suggestion
                    log.debug("received config for check '" + checkName + "'");
```

PR #231, opened 2019-06-03
Fix JMX metric name for G1GC old generation statistics.

### LINKED ISSUE #230: Incorrect name for G1GC old generation JMX metric
The JMX metric name for collecting info on G1GC old generation counts and durations is incorrectly named. The default config file uses the name 'G1 Mixed Generation' when it should be 'G1 Old Generation'. I've confirmed this using visualvm on a running Java 8, 11 and 12 VM.

In the default config file (new-gc-default-jmx-metrics.yaml) the stanza at the bottom that reads

```yaml
- include:
    domain: java.lang
    type: GarbageCollector
    name: G1 Mixed Generation
    attribute:
      CollectionCount:
        alias: jvm.gc.major_collection_count
        metric_type: counter
      CollectionTime:
        alias: jvm.gc.major_collection_time
        metric_type: counter
```
should be
```yaml
- include:
    domain: java.lang
    type: GarbageCollector
    name: G1 Old Generation
    attribute:
      CollectionCount:
        alias: jvm.gc.major_collection_count
        metric_type: counter
      CollectionTime:
        alias: jvm.gc.major_collection_time
        metric_type: counter
```

dpratt:
Fixes #230

### DIFF
--- a/src/main/resources/org/datadog/jmxfetch/new-gc-default-jmx-metrics.yaml
+++ b/src/main/resources/org/datadog/jmxfetch/new-gc-default-jmx-metrics.yaml
@@ -82,6 +82,17 @@
 - include:
     domain: java.lang
     type: GarbageCollector
+    name: G1 Mixed Generation
+    attribute:
+      CollectionCount:
+        alias: jvm.gc.major_collection_count
+        metric_type: counter
+      CollectionTime:
+        alias: jvm.gc.major_collection_time
+        metric_type: counter
+- include:
+    domain: java.lang
+    type: GarbageCollector
     name: G1 Old Generation
     attribute:
       CollectionCount:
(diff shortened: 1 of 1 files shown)


albertvaka:
Hi @dpratt! Thanks for your PR 😃 

Can we be sure `G1 Mixed Generation` is not used in any JVM?

dpratt:
@albertvaka I can't say for sure that it isn't, but across OpenJDK Java 8, 9, 10, 11 and 12 it's referenced as "G1 Old Generation".

I heavily suspect that if it were to show up it would likely be mutually exclusive with the name referenced above - it couldn't hurt to have both of them in the default JMX config file for safety purposes.

albertvaka:
@dpratt can you change the PR to keep both the `Mixed` and `Old` names? It shouldn't hurt to have both, and it gives us the certainty that this is not a breaking change.

dpratt:
Done - I've pushed that change.

albertvaka:
Thanks a lot @dpratt, merging! 😄

PR #237, opened 2019-07-03
Enable launch of jmxfetch app's executors as daemon, default to NO-daemon

labbati:
When embedding the jmxfetch app in dd-trce-java we noticed that even if
we run it in a thread marked as daemon, the jmxfetch app was not terminating
when the main method exited.

After some investigation we noticed that threads in tasks executors were
keeping the thread alive.

This PR DOES NOT change the default behavior, were executors are run as
non-daemons. Adds the possibility, though, to be configured to be
executed as daemon.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -48,6 +48,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Matcher;
@@ -89,12 +90,12 @@ public class App {
         this.appConfig = appConfig;
 
         ExecutorService collectionThreadPool =
-                Executors.newFixedThreadPool(appConfig.getThreadPoolSize());
+                buildExecutorService(appConfig.getThreadPoolSize());
         collectionProcessor =
                 new TaskProcessor(collectionThreadPool, appConfig.getReporter());
 
         ExecutorService recoveryThreadPool =
-                Executors.newFixedThreadPool(appConfig.getReconnectionThreadPoolSize());
+                buildExecutorService(appConfig.getReconnectionThreadPoolSize());
         recoveryProcessor = new TaskProcessor(recoveryThreadPool, appConfig.getReporter());
 
         // setup client
@@ -263,7 +264,7 @@ public class App {
                         + "previous one hogging threads");
                 recoveryProcessor.stop();
                 recoveryProcessor.setThreadPoolExecutor(
-                        Executors.newFixedThreadPool(appConfig.getReconnectionThreadPoolSize()));
+                        buildExecutorService(appConfig.getReconnectionThreadPoolSize()));
             }
 
             List<TaskStatusHandler> statuses =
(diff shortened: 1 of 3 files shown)


src/main/java/org/datadog/jmxfetch/tasks/TaskProcessor.java
```diff
@@ -37,7 +37,7 @@ public void setThreadPoolExecutor(ExecutorService executor) {
      * */
     public boolean ready() {
         ThreadPoolExecutor tpe = (ThreadPoolExecutor) threadPoolExecutor;
-        return !(tpe.getMaximumPoolSize() == tpe.getActiveCount());
+        return !tpe.isTerminated() && !(tpe.getMaximumPoolSize() == tpe.getActiveCount());
```
olivielpeau: just a question: is this change needed as well for the purposes of the PR, or is it an "unrelated" improvement?

src/main/java/org/datadog/jmxfetch/AppConfig.java
```diff
@@ -207,6 +207,13 @@
     @Builder.Default
     private boolean targetDirectInstances = false;
 
+    /**
+     * Boolean setting to determine whether to ignore jvm_direct instances.
+     * If set to true, other instances will be ignored.
```
olivielpeau: This inline comment needs to be changed.

I think it'd be useful for future reference to explain the use cases where we want to enable the option (namely, when JMXFetch is embedded in a client app, typically for the java tracer, so that the client app's exit doesn't block on the termination of these internal threads)

src/main/java/org/datadog/jmxfetch/tasks/TaskProcessor.java
labbati: This is required because when running as daemon the underlying thread pull executor can be terminated before the shutdown hook. The problem is that starting from the shutdown hook we we check is a thread executor is ready and if not ready we recreate.

I had a false positive (ready) when actually it was terminated.

I did not had a separate method as looking at the code around I thought that adding this check tot he ready function made sense.

Totally open to that, though.

src/main/java/org/datadog/jmxfetch/AppConfig.java
labbati: Totally agree, this is a left over after copying and pasting another property. Changing right away.

src/main/java/org/datadog/jmxfetch/tasks/TaskProcessor.java
olivielpeau: thanks for the explanation!

Hmm, looking at where the `ready` function is used, and the case you described: does adding the condition here actually solve the issue you mentioned?

i.e.: if the underlying thread pull executor is terminated before the shutdown hook is run, the `ready` function here would return `false`, so the logic in the `App` would recreate a new thread pool executor. Is that the desired behavior?

labbati: Thanks for looking into this. This is the expected behavior based on my investigation. The thread pool is recreated (to do some final work) and then once that work is done it will automatically exit.

I tested both in web apps (i.e. long running processes) and short lived CLI apps and this is how I could observe it behaves.

Anyway this ring a bell in you I can look better into this. Do you have a specific concern I can specifically look for?

labbati: Hi @olivielpeau, does my answer make sense to you? Please let me know

olivielpeau: Ok, thanks for the explanation, makes sense!

labbati:
Thanks a lot everyone! Should I merge to master myself? Also, can we plan a release to include this fix?

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -290,6 +291,24 @@ public TaskStatusHandler invoke(
...
+     * and executed as daemons if {@link AppConfig#isDaemon()} is true. Defaults to false.
+     *
+     * @param size The thread pool size
+     * @return The create executor
+     */
+    private ExecutorService buildExecutorService(int size) {
+        return Executors.newFixedThreadPool(size, new ThreadFactory() {
+            @Override
+            public Thread newThread(Runnable runnable) {
+                Thread thread = Executors.defaultThreadFactory().newThread(runnable);
```
tylerbenson: I would suggest assigning a proper name to the thread... this would have greatly helped our troubleshooting.  Maybe next time someone does some work here?

PR #240, opened 2019-07-22
[deps] Update jackson-databind to 2.9.9.3

olivielpeau:
Updates `jackson-databind` to 2.9.9.3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
         <jcommander.version>1.35</jcommander.version>
         <log4j.version>1.2.17</log4j.version>
         <jackson.version>2.9.9</jackson.version>
-        <jackson-databind.version>2.9.9.2</jackson-databind.version>
+        <jackson-databind.version>2.9.9.3</jackson-databind.version>
         <snakeyaml.version>1.13</snakeyaml.version>
         <!-- Note: using project checkstyle with AOSP style
         find the default google java checkstyle config here -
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -43,6 +43,7 @@
         <jcommander.version>1.35</jcommander.version>
         <log4j.version>1.2.17</log4j.version>
         <jackson.version>2.9.9</jackson.version>
+        <jackson-databind.version>2.9.9.1</jackson-databind.version>
```
tylerbenson: Why have a discrepancy between `jackson.version` and `jackson-databind.version`?

olivielpeau: 2.9.9.1 and 2.9.9.2 are "micro-patches" that were only released for `jackson-databind`, see https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9#micro-patches

pom.xml
```diff
@@ -43,6 +43,7 @@
         <jcommander.version>1.35</jcommander.version>
         <log4j.version>1.2.17</log4j.version>
         <jackson.version>2.9.9</jackson.version>
+        <jackson-databind.version>2.9.9.2</jackson-databind.version>
```
KSerrania:
```suggestion
        <jackson-databind.version>2.9.9.3</jackson-databind.version>
```
Not sure at all we're affected by this, but `2.9.9.2` introduced a regression that was fixed today in `2.9.9.3` according to [this](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9#micro-patches), maybe we can include it as well while we're updating?

olivielpeau: nice find, let me update this then (here's hoping there won't be yet another micro release in a few days 😅 )

olivielpeau: done, can you review again @KSerrania?

PR #250, opened 2019-09-27
Bump jackson-core to v2.10.0

olivielpeau:


arbll:
Any reason not to use 2.9.10 instead ?

olivielpeau:
Not really, except that 2.10.0 > 2.9.10 so I was thinking we might as well bump to the latest.

arbll:
Does not seem to bring anything breaking, feel free to merge since it passes CI

PR #261, opened 2020-02-04
Fix ThreadPool creation

jpbempel:
`Executors.defaultThreadFactory()` is creating a new `ThreadFactory` for each new thread created, messing up thread names and pool belongings.
Now, we are passing a pool name when building a new executor service, and handling counter for
threads belonging to this thread pool

Thread names before:
pool-2-thread-1
pool-3-thread-1
pool-4-thread-1
pool-5-thread-1

Thread names after:
jmxfetch-collectionPool-1
jmxfetch-collectionPool-2
jmxfetch-collectionPool-3
jmxfetch-recoveryPool-1

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -69,8 +69,8 @@ public class App {
     private static final int AD_MAX_MAG_INSTANCES =
             4; // 1000 instances ought to be enough for anyone
     private static final Charset UTF_8 = Charset.forName("UTF-8");
-    private static final String COLLECTION_POOL_NAME = "collectionPool";
-    private static final String RECOVERY_POOL_NAME = "recoveryPool";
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -68,6 +69,8 @@
     private static final int AD_MAX_MAG_INSTANCES =
             4; // 1000 instances ought to be enough for anyone
     private static final Charset UTF_8 = Charset.forName("UTF-8");
+    private static final String COLLECTION_POOL_NAME = "collectionPool";
+    private static final String RECOVERY_POOL_NAME = "recoveryPool";
```
tylerbenson: consider prefixing with `jmxfetch-`

jpbempel: done

PR #262, opened 2020-02-05
Add Embedded mode

jpbempel:
When JMXFetch is embedded into the client app through tracer agent for example, we want to have a special mode
to avoid creating thread pools that are only used for dealing with JMX remote connections
TaskProcessor is taking care of the case where there is no ThreadPoolExecutor and calling tasks by the calling thread

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -92,15 +92,16 @@ public class App {
     public App(AppConfig appConfig) {
         this.appConfig = appConfig;
 
-        ExecutorService collectionThreadPool =
-                buildExecutorService(appConfig.getThreadPoolSize(), COLLECTION_POOL_NAME);
-        collectionProcessor =
-                new TaskProcessor(collectionThreadPool, appConfig.getReporter());
-
-        ExecutorService recoveryThreadPool =
-                buildExecutorService(appConfig.getReconnectionThreadPoolSize(), RECOVERY_POOL_NAME);
+        ExecutorService collectionThreadPool = null;
+        ExecutorService recoveryThreadPool = null;
+        if (!appConfig.isEmbedded()) { // Creates executors in standalone mode only
+            collectionThreadPool = buildExecutorService(appConfig.getThreadPoolSize(),
+                    COLLECTION_POOL_NAME);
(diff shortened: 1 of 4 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -89,15 +89,14 @@
...
-                buildExecutorService(appConfig.getReconnectionThreadPoolSize());
+        ExecutorService collectionThreadPool = null;
+        ExecutorService recoveryThreadPool = null;
+        if (!appConfig.isEmbedded()) { // Creates executors in standalone mode only
+            collectionThreadPool = buildExecutorService(appConfig.getThreadPoolSize());
+            recoveryThreadPool = buildExecutorService(appConfig.getReconnectionThreadPoolSize());
+        }
         recoveryProcessor = new TaskProcessor(recoveryThreadPool, appConfig.getReporter());
-
+        collectionProcessor = new TaskProcessor(collectionThreadPool, appConfig.getReporter());
```
tylerbenson: instead of an embedded setting, perhaps it would be better to take an ExecutorService as an option?

What is the impact if it doesn't have these threadpools?

jpbempel: I am not sure to follow your point here.
In embedded mode, collectionThreadPool & recoveryThreadPool variables are null, so TaskProcessor instances don't have any ExecutorService.

tylerbenson: Then what thread is jmxfetch even running on?

jpbempel: The calling thread, i.e. the thread calling `App.run()` method.
In case of the tracer, this is the dd-jmx-collector thread created by the tracer.
see: https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-jmxfetch/src/main/java/datadog/trace/agent/jmxfetch/JMXFetch.java#L90

src/test/java/org/datadog/jmxfetch/tasks/TestTaskProcessor.java
```diff
@@ -155,4 +145,32 @@ public TaskStatusHandler invoke(Instance instance, Future<Boolean> future, Repor
             }
         }
     }
+
+    @Test
+    public void embeddedTaskProcessor() throws Throwable {
```
truthbk: 🍰

PR #264, opened 2020-02-07
add entity-id by default to the dogstatsdclient

clamoriniere:
Create the StatsDClient with "entity-id" set to "none" to avoid having dogstatsd server adding origin tags, when the connection is done with UDS.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
@@ -35,6 +35,10 @@ public class StatsdReporter extends Reporter {
 
     private void init() {
         initializationTime = System.currentTimeMillis();
+
+        // Only set the entityID to "none" if UDS communication is activated
+        String entityID = this.statsdPort == 0 ? "none" : null;
+
         /* Create the StatsDClient with "entity-id" set to "none" to avoid
            having dogstatsd server adding origin tags, when the connection is
            done with UDS. */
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -35,13 +35,18 @@ public StatsdReporter(String statsdHost, int statsdPort) {
...
         initializationTime = System.currentTimeMillis();
+        /* Create the StatsDClient with "entity-id" set to "none" to avoid
+           having dogstatsd server adding origin tags, when the connection is
+           done with UDS. */
         statsDClient =
                 new NonBlockingStatsDClient(
                         null,
                         this.statsdHost,
                         this.statsdPort,
+                        Integer.MAX_VALUE,
```
truthbk: reference note: `Integer.MAX_VALUE` is the default queue length, so this will *NOT* change any behavior.

src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -35,13 +35,18 @@ public StatsdReporter(String statsdHost, int statsdPort) {
...
                         new String[] {},
-                        new LoggingErrorHandler());
+                        new LoggingErrorHandler(),
+                        "none");
```
truthbk: shouldn't we only be doing this if `this.statsdPort == 0`, otherwise this will also apply to UDP connections.

clamoriniere: good catch! I'm updating the PR

clamoriniere:
@truthbk could you do a second review? 
I added the `this.statsdPort == 0` test.

truthbk:
CI issue is unrelated, will fix in a separate PR.

PR #266, opened 2020-02-11
Replace LinkedHashMap by HashMap or Map

jpbempel:
Configurations come mostly as LinkedHashMap from Yaml or Json parsers
But usage inside JMXFetch is mostly as HashMap.
As LinkedHashMap is heavier regarding memory footprint, we prefer to
avoid its usage by just replacing by HashMap. Also it does not seem
That order matters in the case of JMXFetch usage.
I have changed also the type to use Map interface as it will be easier
to change the implementation if required in the future without
changing this type everywhere.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -35,7 +35,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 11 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -849,16 +848,16 @@ public void init(boolean forceNewConnection) {
                 it.remove();
             }
 
-            ArrayList<LinkedHashMap<String, Object>> configInstances =
-                    ((ArrayList<LinkedHashMap<String, Object>>) yamlConfig.getYamlInstances());
+            ArrayList<Map<String, Object>> configInstances =
```
devinsba: I know this is outside the scope of your changes, and is more a style thing, but since we are changing to the interface with `Map` maybe we should do the same with `List`?

jpbempel: This is effectively outside of the scope and it will be addressed in a next PR :)
I think there are already enough changes for a PR 😅

PR #267, opened 2020-02-12
Add a new command to test rate metrics

therve:
For end to end testing, we use the `list_with_metrics` command. It only
does one iteration though, which means we don't have output for counters
and rates. This adds a new command running 2 iterations.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/JsonReporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/JsonReporter.java
@@ -8,6 +8,7 @@ import org.datadog.jmxfetch.Instance;
 import org.datadog.jmxfetch.JmxAttribute;
 
 import java.io.IOException;
+import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


src/main/java/org/datadog/jmxfetch/reporter/JsonReporter.java
```diff
@@ -49,11 +51,14 @@ public void doSendServiceCheck(String checkName, String status, String message,
         System.out.println("=== JSON ===");
         ObjectMapper mapper = new ObjectMapper();
         mapper.enable(SerializationFeature.INDENT_OUTPUT);
+        StringWriter writer = new StringWriter();
```
truthbk: 👍

therve:
> Makes me wonder if we shouldn't really keep a single command (ie. `ACTION_LIST_WITH_METRICS`) and just make sure _that_ command returns all metrics, including rates. Do you see value in having two separate commands?

Ideally we'd have a command with a parameter flag. If we change the current we'll break our end to end tests. It may not be a huge deal, but in regular tests we like having the distinction between rate and non-rate.

PR #268, opened 2020-02-12
Replace LinkedList by ArrayList and or List

jpbempel:
LinkedList are heavier regarding memory footprint, we prefer to void
its usage by just replacing by ArrayList.
I have changed also the type to use List interface as it will be
easier to change implementation if required.
Some lists are also pre-sized when possible.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Filter.java
+++ b/src/main/java/org/datadog/jmxfetch/Filter.java
@@ -1,15 +1,18 @@
 package org.datadog.jmxfetch;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Set;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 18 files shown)


src/main/java/org/datadog/jmxfetch/Filter.java
```diff
@@ -49,20 +52,15 @@ public String toString() {
...
-        } catch (ClassCastException e) {
-            return new ArrayList<String>() {
-                {
-                    add(((String) toCast));
-                }
-            };
+        if (toCast instanceof List) {
+            return (List<String>) toCast;
         }
+        return Collections.singletonList((String) toCast);
```
devinsba: I just want to make sure the result of this call is not expected to be modified anywhere. I see one assignment to a instance variable `this.excludeTags`

jpbempel: You're right, there is some add operations on it afterward

PR #269, opened 2020-02-14
Support sending a `service` tag

ofek:
We want to feature the notion of service more prominently for integrations so that we can correlate metrics with logs and traces

See:

https://github.com/DataDog/datadog-agent/pull/4877
https://github.com/DataDog/datadog-agent/pull/4895

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -88,6 +88,7 @@ public class Instance {
     private Map<String, Object> instanceMap;
     private Map<String, Object> initConfig;
     private String instanceName;
+    private String service;
     private Map<String, String> tags;
     private String checkName;
     private int maxReturnedMetrics;
@@ -138,6 +139,14 @@ public class Instance {
         } else {
             // Allow global overrides
             this.refreshBeansPeriod = appConfig.getRefreshBeansPeriod();
+        }
+
+        this.service = (String) instanceMap.get("service");
+        if (this.service == null && initConfig != null) {
+            this.service = (String) initConfig.get("service");
+        }
+        if (this.service != null && this.service != "") {
+            this.tags.put("service", this.service);
         }
 
         this.minCollectionPeriod = (Integer) instanceMap.get("min_collection_interval");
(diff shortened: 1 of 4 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -140,6 +141,14 @@ public Instance(
             this.refreshBeansPeriod = appConfig.getRefreshBeansPeriod();
         }
 
+        this.service = (String) instanceMap.get("service");
+        if (this.service == null && initConfig != null) {
```
devinsba: Since you check for empty string below do you want to do that here too?

ofek: done!

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -140,6 +141,14 @@ public Instance(
             this.refreshBeansPeriod = appConfig.getRefreshBeansPeriod();
         }
 
+        this.service = (String) instanceMap.get("service");
+        if ((this.service == null || this.service == "") && initConfig != null) {
```
remeh: I think it'll be better to do:
```java
        if ((this.service == null || this.service.isEmpty()) && initConfig != null) {
```
You may already know but you should not use `==` with string in Java, if you don't want to go with `.isEmpty` you should use `.equals`

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -140,6 +141,14 @@ public Instance(
...
+            this.service = (String) initConfig.get("service");
+        }
+        if (this.service != null && this.service != "") {
```
remeh: `this.service != null && !this.service.isEmpty()`

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -667,6 +676,10 @@ private void refreshBeansList() throws IOException {
         }
         tags.add("instance:" + this.instanceName);
 
+        if (this.service != null && this.service != "") {
```
remeh: `this.service != null && !this.service.isEmpty()`

PR #271, opened 2020-02-14
Replace Map metric by a Metric class

jpbempel:
During collection process, metrics are collected as
Map<String, Object>, but keys are known and fixed so we can replace
this map to a POJO which will be lighter in term of memory footprint
and allocation.


truthbk:
Going to merge this. Please feel free to address any of the feedback, which didn't affect any of the  logic introduced here, in a follow-up PR. Thank you! 🙇

PR #273, opened 2020-02-17
Cache Metric instances

jpbempel:
Use Metric class as a cache and avoid recreating it for each
collection. Only the value is updated. For simple & complex attribute
Tags are also cached into the Metric instance.
For complex & tabular attribute, cache is done per sub attribute into
a new JMXSubAttribute abstract class
Transfer checkName into the attribute to be inserted into the Metric
instance at creation


src/main/java/org/datadog/jmxfetch/JmxTabularAttribute.java
```diff
@@ -133,10 +135,12 @@ private void populateSubAttributeList(Object value) {
             String dataKey = entry.getKey();
             List<String> subSub = entry.getValue();
             for (String metricKey : subSub) {
-                String alias = convertMetricName(getAlias(metricKey));
+                String alias = getAlias(metricKey);
```
truthbk: 🍰 this should be fine because `getAlias()` already calls `convertMetricName`, we were just duplicating work.

PR #277, opened 2020-02-25
Add class name and class regex filters

AlexandreYang:
## What this PR does

Add class name and class regex filters

## Motivation

Allow creating rules based on class name filters.

For example Kafka / Confluent Platform uses yammer metrics that provide different MBean classes like `Gauge`, `Histogram`, `Timer`, `Meter`. Since each class have different attributes and require specific rules, a filter by class can be helpful in targeting all metrics from a specific MBean class.

The classes are like follow:

```
- com.yammer.metrics.reporting.JmxReporter$Gauge
- com.yammer.metrics.reporting.JmxReporter$Histogram
- com.yammer.metrics.reporting.JmxReporter$Timer
- com.yammer.metrics.reporting.JmxReporter$Meter
```

That will allow creating rules tailored to each class:

```yaml
jmx_metrics:
  - include:
      domain_regex: kafka\..*
      class_regex: .*$Gauge
      attribute:
        Value:
          alias: $domain.$type.$name
          metric_type: gauge
  - include:
      domain_regex: kafka\..*
      class_regex: .*$Histogram
      attribute:
        Count:
          alias: $domain.$type.$name.count
          metric_type: count
        Mean:
          alias: $domain.$type.$name.avg
          metric_type: gauge
        99thPercentile:
          alias: $domain.$type.$name.99percentile
          metric_type: gauge

```

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -495,7 +495,7 @@ public class Instance {
             String className;
 
             try {
-                className = connection.getClassForBean(beanName);
+                className = connection.getClassNameForBean(beanName);
                 log.debug("ClassName: " + className);
             } catch (Exception e) {
                 log.warn("Cannot get class name " + e.getMessage());
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -20,7 +20,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import javax.management.*;
+import javax.management.MBeanAttributeInfo;
+import javax.management.ObjectName;
 import javax.security.auth.login.FailedLoginException;
 
 @Slf4j
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -21,8 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import javax.management.MBeanAttributeInfo;
-import javax.management.ObjectName;
+import javax.management.*;
 import javax.security.auth.login.FailedLoginException;
 
 @Slf4j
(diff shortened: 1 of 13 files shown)


src/test/java/org/datadog/jmxfetch/SimpleTestJavaApp2.java
```diff
@@ -0,0 +1,5 @@
+package org.datadog.jmxfetch;
+
+public class SimpleTestJavaApp2 extends SimpleTestJavaApp {
```
truthbk: Not sure we need this in a new file, I think it'll be better if you use an inner class in `TestApp.java`.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -478,21 +478,24 @@ private void getMatchingAttributes() throws IOException {
                     break;
                 }
             }
+            String className;
             MBeanAttributeInfo[] attributeInfos;
-
             try {
+                log.debug("Getting class name for bean: " + beanName);
```
truthbk: note to self: I was going to say these new log lines perhaps didn't add much value, but since we're calling methods on `connection` we can probably trigger an `IOException` with these, so it's probably justified.

src/test/java/org/datadog/jmxfetch/TestApp.java
```diff
@@ -256,6 +256,61 @@ public void testDomainRegex() throws Exception {
         assertEquals(15, metrics.size());
     }
 
+    @Test
+    public void testClassInclude() throws Exception {
```
truthbk: 🍰

src/test/java/org/datadog/jmxfetch/SimpleTestJavaApp2.java
AlexandreYang: Yes, the initial reason to use a new file was to have a simpler name to match. But it's it's probably fine to use a inner class too. Changed in my last commit.

Using a separate class:
```
class: org.datadog.jmxfetch.SimpleTestJavaApp2
```

Using a inner class:
```
class: org.datadog.jmxfetch.TestApp$1SimpleTestJavaAnotherApp
```

PR #284, opened 2020-03-16
Add check prefix config

AlexandreYang:
#### Add option to configure the prefix used for service checks.

This solves two issues:

- Some integration use prefix that doesn't necessarily correspond to the integration name.
  - Confluent Platform use `confluent.` as prefix. https://github.com/DataDog/integrations-core/pull/5733/files#diff-9d75daf4238a279ff75c4a0c591b3d54R6
- When integration is a multi word e.g. `my_integration`, the `_` is currently is stripped out. Which is usually not wanted.
  - example of issue https://github.com/DataDog/integrations-core/pull/5999

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -90,6 +90,7 @@ public class Instance {
     private String service;
     private Map<String, String> tags;
     private String checkName;
+    private String checkPrefix;
     private int maxReturnedMetrics;
     private boolean limitReached;
     private Connection connection;
@@ -184,6 +185,10 @@ public class Instance {
                                 + "Please define a name in your instance configuration");
                 this.instanceName = this.checkName;
             }
+        }
+
+        if (initConfig != null) {
+            this.checkPrefix = (String) initConfig.get("check_prefix");
         }
 
         // Alternative aliasing for CASSANDRA-4009 metrics
@@ -704,6 +709,11 @@ public class Instance {
     /** Returns the check name. */
     public String getCheckName() {
         return this.checkName;
+    }
+
+    /** Returns the check prefix. */
+    public String getCheckPrefix() {
+        return this.checkPrefix;
     }
 
     /** Returns the maximum number of metrics an instance may collect. */
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -90,7 +90,7 @@ public class Instance {
     private String service;
     private Map<String, String> tags;
     private String checkName;
-    private String checkPrefix;
+    private String serviceCheckPrefix;
     private int maxReturnedMetrics;
     private boolean limitReached;
     private Connection connection;
@@ -188,7 +188,7 @@ public class Instance {
         }
 
         if (initConfig != null) {
-            this.checkPrefix = (String) initConfig.get("check_prefix");
+            this.serviceCheckPrefix = (String) initConfig.get("service_check_prefix");
         }
 
         // Alternative aliasing for CASSANDRA-4009 metrics
@@ -712,8 +712,8 @@ public class Instance {
     }
 
     /** Returns the check prefix. */
-    public String getCheckPrefix() {
-        return this.checkPrefix;
+    public String getServiceCheckPrefix() {
+        return this.serviceCheckPrefix;
     }
 
     /** Returns the maximum number of metrics an instance may collect. */
--- a/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
@@ -156,7 +156,8 @@ public abstract class Reporter {
     }
 
     /** Submits service check. */
-    public void sendServiceCheck(String checkName, String serviceCheckName, String status, String message, String[] tags) {
+    public void sendServiceCheck(String checkName, String serviceCheckName,
+                                 String status, String message, String[] tags) {
         this.incrementServiceCheckCount(checkName);
         this.doSendServiceCheck(serviceCheckName, status, message, tags);
     }
--- a/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
@@ -2,7 +2,6 @@ package org.datadog.jmxfetch.reporter;
 
 import com.timgroup.statsd.ServiceCheck;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang.StringUtils;
 
 import org.datadog.jmxfetch.App;
 import org.datadog.jmxfetch.Instance;
@@ -157,11 +156,8 @@ public abstract class Reporter {
     }
 
     /** Submits service check. */
-    public void sendServiceCheck(String checkName, String status, String message, String[] tags) {
+    public void sendServiceCheck(String checkName, String serviceCheckName, String status, String message, String[] tags) {
         this.incrementServiceCheckCount(checkName);
-        String serviceCheckName = String.format(
-            "%s.can_connect", Reporter.formatServiceCheckPrefix(checkName));
-
         this.doSendServiceCheck(serviceCheckName, status, message, tags);
     }
 
@@ -182,13 +178,6 @@ public abstract class Reporter {
 
     protected Map<String, Integer> getServiceCheckCountMap() {
         return this.serviceCheckCount;
-    }
-
-    /** Formats the service check prefix. */
-    public static String formatServiceCheckPrefix(String fullname) {
-        String[] chunks = fullname.split("\\.");
-        chunks[0] = chunks[0].replaceAll("[A-Z0-9:_\\-]", "");
-        return StringUtils.join(chunks, ".");
     }
 
     protected ServiceCheck.Status statusToServiceCheckStatus(String status) {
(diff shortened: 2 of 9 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -186,6 +187,10 @@ public Instance(
             }
         }
 
+        if (initConfig != null) {
+            this.checkPrefix = (String) initConfig.get("check_prefix");
```
olivielpeau: I'd prefer naming this option `service_check_prefix` to reflect more closely what it does, since the prefix is only applied to the service check

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -157,11 +156,9 @@ public void sendMetrics(List<Metric> metrics, String instanceName, boolean canon
     }
 
     /** Submits service check. */
-    public void sendServiceCheck(String checkName, String status, String message, String[] tags) {
+    public void sendServiceCheck(String checkName, String serviceCheckName,
+                                 String status, String message, String[] tags) {
```
olivielpeau: I think we can/should remove `checkName` from the signature:

```suggestion
    public void sendServiceCheck(String serviceCheckName, String status,
                                 String message, String[] tags) {
```

(`incrementServiceCheckCount` seems to be only used to increment an internal counter only used in tests, let's make it track the full service check names instead of simply the check names, and update the tests accordingly)

src/main/java/org/datadog/jmxfetch/util/ServiceCheckHelper.java
```diff
@@ -0,0 +1,12 @@
+package org.datadog.jmxfetch.util;
+
+import org.apache.commons.lang.StringUtils;
+
+public class ServiceCheckHelper {
+    /** Formats the service check prefix. */
```
olivielpeau: Let's add an explanation here about what this formatting does, and why, see explanation here: https://github.com/DataDog/jmxfetch/commit/0428c41ebf7a14404ae50928e3ecfc229701c042

This was, in hindsight, a poor solution to the technical problem at the time. As a way to start the transition to not using this formatting logic anymore, maybe it'd make sense to send both the formatted service check name and the unformatted one when your new option `check_prefix` is not specified, what do you think?

src/main/java/org/datadog/jmxfetch/util/ServiceCheckHelper.java
```diff
@@ -0,0 +1,12 @@
...
+    public static String formatServiceCheckPrefix(String fullname) {
```
olivielpeau: is this function tested anywhere? (if not, we should add a test for it to improve test coverage a bit)

src/main/java/org/datadog/jmxfetch/Instance.java
AlexandreYang: I was thinking it might be also used later for metrics if there is a need for that later. That's why I used `check_prefix` instead of `service_check_prefix`, WDYT?

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
AlexandreYang: The service check count is also used for reporting here: https://github.com/DataDog/jmxfetch/blob/a4a17db0a778ee5cdccf290b892db8a81812b655/src/main/java/org/datadog/jmxfetch/App.java#L808 

It seems the intent of the reported service check count is about `the total of service checks sent for a specific check name` (not only about `X.can_connect` service check). So, if we have more service checks sent later, it seems to make sense to count by `check name` as it's currently done.

src/main/java/org/datadog/jmxfetch/util/ServiceCheckHelper.java
AlexandreYang: > Formats the service check prefix name to enable having versionned .yaml
configuration files for the same checks (ex. : activemq/activemq_58)
while keeping on sending the same metric prefix. All numbers,
underscores and some other characters are simply removed from the prefix
if they are present.

I'm still not sure to understand why there was a need to strip those characters based on this comment ☝️ in order to have versioning (I don't have much context on this versioning concept).

> This was, in hindsight, a poor solution to the technical problem at the time. As a way to start the transition to not using this formatting logic anymore, maybe it'd make sense to send both the formatted service check name and the unformatted one when your new option check_prefix is not specified, what do you think?

Yes, make sense. So, if the check_name is `foo_bar`:

- If `check_prefix = foo`, we send 2 service checks `foo.can_connect` and `foobar.can_connect`
- If `check_prefix` is not set, we send 2 service checks `foo_can.connect` and `foobar.can_connect`
- We deprecate the current formatting: `foobar.can_connect`

src/main/java/org/datadog/jmxfetch/util/ServiceCheckHelper.java
AlexandreYang: Sure, added a test for it.

src/main/java/org/datadog/jmxfetch/Instance.java
olivielpeau: If we want to use such a prefix for metrics as well in the future, I think it'll be easier for backward compatibility to define a new config option (that would at that point be called `integration_prefix` for example) and apply it to both service checks and metrics; and still support `service_check_prefix` for service checks only.

src/main/java/org/datadog/jmxfetch/util/ServiceCheckHelper.java
olivielpeau: > I'm still not sure to understand why there was a need to strip those characters based on this comment ☝️ in order to have versioning (I don't have much context on this versioning concept).

at the time there were 2 JMXFetch integrations for `activemq`: one called `activemq` for older versions of activemq, the other called `activemq_58` for v5.8+ of activemq, see https://github.com/DataDog/dd-agent/tree/5.10.x/conf.d

And we still wanted both integrations to send the service check with the `activemq` prefix, hence this formatting function...

> So, if the check_name is foo_bar:
> - If check_prefix = foo, we send 2 service checks foo.can_connect and foobar.can_connect
> - If check_prefix is not set, we send 2 service checks foo_can.connect and foobar.can_connect
> - We deprecate the current formatting: foobar.can_connect

I agree overall. In the case of check_prefix = foo, it's up to you whether we still want to send `foobar.can_connect` (we should if you need that behavior for existing integrations where you'll start using `check_prefix`, otherwise we should only send `foo.can_connect`)

> But then we need a deprecation process for when we should remove the old behaviour ?

Yes, unclear how we'll do that. But at least JMXFetch will already be sending the "right" service check name.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
olivielpeau: oh, good catch, hadn't seen its use in `App.java`.

That said, it looks like the service check count is then set on the stats of the instance, not the check as a whole (see https://github.com/DataDog/jmxfetch/blob/37fc55f929f897066014e89946e666f03ee27eff/src/main/java/org/datadog/jmxfetch/Status.java#L104), so it's better to actually report the number of service checks sent for that specific instance of the check

src/main/java/org/datadog/jmxfetch/Instance.java
AlexandreYang: Sure, make sense. Changed to `service_check_prefix` instead.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
AlexandreYang: Seems there is a bug here, looks like the `service_check_count` is always reporting 0:

````

========
JMXFetch
========

  Initialized checks
  ==================
    confluent_platform
      instance_name : confluent_platform-localhost-31000
      message : <no value>
      metric_count : 26
      service_check_count : 0
      status : OK
      instance_name : confluent_platform-localhost-31001
      message : <no value>
      metric_count : 312
      service_check_count : 0
      status : OK
      instance_name : confluent_platform-localhost-31002
      message : <no value>
      metric_count : 205
      service_check_count : 0
      status : OK
      instance_name : confluent_platform-localhost-31008
      message : <no value>
      metric_count : 205
      service_check_count : 0
      status : OK
      instance_name : confluent_platform-localhost-31006
      message : <no value>
      metric_count : 29
      service_check_count : 0
      status : OK
      instance_name : confluent_platform-localhost-31007
      message : <no value>
      metric_count : 91
      service_check_count : 0
      status : OK
      instance_name : confluent_platform-localhost-31004
      message : Unable to instantiate or initialize instance localhost:31004. Is the target JMX Server or JVM running? java.io.IOException: Failed to retrieve RMIServer stub: javax.naming.ServiceUnavailableException [Root exception is java.rmi.ConnectException: Connection refused to host: localhost; nested exception is:
	java.net.ConnectException: Connection refused (Connection refused)]
      metric_count : 0
      service_check_count : 0
      status : ERROR
```

> That said, it looks like the service check count is then set on the stats of the instance, not the check as a whole 

Agree this that. Do you mind if we adresse that in a separate PR (and the potential bug mentioned above)? It seems out of the scope of this PR.

src/main/java/org/datadog/jmxfetch/util/ServiceCheckHelper.java
AlexandreYang: Just made changes related to this. Let me know if it's ok for you.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
olivielpeau: hmm, thanks for noticing this. Let's address this separately then agreed, thanks for adding a card with context in our backlog

PR #282, opened 2020-03-10
Use custom RMISocketFactory to specify socket connect timeout.

sirianni:
Fixes #279

Issue #279, opened 2020-03-04
Allow RMI socket connection timeout to be configurable


sirianni:
## Problem
We are seeing issues where `jmxfetch` is having prolonged collection cycles when monitored apps are removed or restarted.

For example, note the 68 second collection cycle when this kafka broker is restarted.
```
 2020-03-04 16:00:48,639 | DEBUG| Reporter | Instance kafka-10.5.225.11-7203 is sending 281 metrics to the metrics reporter during collection #152
 2020-03-04 16:00:49,493 | DEBUG| App | Iteration ran in 964 ms
 2020-03-04 16:01:04,640 | DEBUG| Reporter | Instance kafka-10.5.225.11-7203 is sending 281 metrics to the metrics reporter during collection #153
 2020-03-04 16:01:04,652 | DEBUG| App | Iteration ran in 140 ms
 2020-03-04 16:02:27,753 | DEBUG| App | Iteration ran in 68081 ms
 2020-03-04 16:02:57,185 | DEBUG| Reporter | Instance kafka-10.5.225.12-7203 is sending 250 metrics to the metrics reporter during collection #155
 2020-03-04 16:02:57,210 | DEBUG| App | Iteration ran in 14406 ms
 2020-03-04 16:03:12,375 | DEBUG| Reporter | Instance kafka-10.5.225.12-7203 is sending 250 metrics to the metrics reporter during collection #156
 2020-03-04 16:03:12,391 | DEBUG| App | Iteration ran in 166 ms
 2020-03-04 16:03:27,547 | DEBUG| Reporter | Instance kafka-10.5.225.12-7203 is sending 250 metrics to the metrics reporter during collection #157
 2020-03-04 16:03:27,555 | DEBUG| App | Iteration ran in 142 ms
```

During the long collection cycle, I noticed the `MetricCollectionTask` hung up on the following stack trace:
```
 "pool-9-thread-1" #31 prio=5 os_prio=0 cpu=2222.15ms elapsed=1233.58s tid=0x00007fb988350000 nid=0x293 runnable  [0x00007fb97090f000]
    java.lang.Thread.State: RUNNABLE
 	at java.net.PlainSocketImpl.socketConnect(java.base@11.0.6-ea/Native Method)
 	at java.net.AbstractPlainSocketImpl.doConnect(java.base@11.0.6-ea/AbstractPlainSocketImpl.java:399)
 	- locked <0x000000066bd3fe00> (a java.net.SocksSocketImpl)
[...]

PR #285, opened 2020-03-24
Fix rmi socket connection timeout

prognant:
Based on #283 & #282 

It aims to provide piece of resolution for https://github.com/DataDog/jmxfetch/issues/279 
It allows to set a RMI connection timeout on a per-instance basis, using the configuration parameter `rmi_connection_timeout` (in milliseconds).

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/RemoteConnection.java
+++ b/src/main/java/org/datadog/jmxfetch/RemoteConnection.java
@@ -10,6 +10,7 @@ import javax.management.remote.JMXConnector;
 import javax.management.remote.JMXServiceURL;
 import javax.management.remote.rmi.RMIConnectorServer;
 import javax.rmi.ssl.SslRMIClientSocketFactory;
+import java.rmi.server.RMISocketFactory;
 
 @Slf4j
 public class RemoteConnection extends Connection {
@@ -88,6 +89,8 @@ public class RemoteConnection extends Connection {
 
         // Set an RMI timeout so we don't get stuck waiting for a bean to report a value
         System.setProperty("sun.rmi.transport.tcp.responseTimeout", rmiTimeout);
+
+        RMISocketFactory.setSocketFactory(new SocketFactory(Integer.parseInt(rmiTimeout)));
 
         createConnection();
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/datadog/jmxfetch/RemoteConnection.java
```diff
@@ -27,6 +29,8 @@
     private static final String KEY_STORE_PASSWORD_KEY = "key_store_password";
     private static final String DEFAULT_RMI_RESPONSE_TIMEOUT =
             "15000"; // Match the collection period default
+    // Match the default RMI connection timeout value
+    private static final Integer DEFAULT_RMI_CONNECTION_TIMEOUT = Integer.valueOf(0);
```
HantingZhang2: Hi @prognant, I'm not sure if we should put 0 here as the default since 0 means infinite and the purpose of this PR is to put a non-zero default value for the socket timeout..

prognant: @HantingZhang2  I agree, I'm always a little bit concerned about changing a default value, in this cas we may use something like 15sec (as initially suggested) or 30sec it should not impact existing deployment.

src/main/java/org/datadog/jmxfetch/RemoteConnection.java
```diff
@@ -89,6 +93,18 @@ public RemoteConnection(Map<String, Object> connectionParams) throws IOException
         // Set an RMI timeout so we don't get stuck waiting for a bean to report a value
         System.setProperty("sun.rmi.transport.tcp.responseTimeout", rmiTimeout);
 
+        Integer rmiConnectionTimeout = DEFAULT_RMI_CONNECTION_TIMEOUT;
+        try {
+            rmiConnectionTimeout = (Integer) connectionParams.get("rmi_connection_timeout");
+        } catch (final ClassCastException e) {
+            rmiConnectionTimeout = 
+                Integer.parseInt((String) connectionParams.get("rmi_connection_timeout"));
```
truthbk: note to self: this would not trigger a `java.lang.NumberFormatException` when `connectionParams.get("rmi_connection_timeout")` is null, because the cast of `(Integer) null` is valid. So we're good.

PR #287, opened 2020-04-01
Fix json service check status

AlexandreYang:


src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -192,6 +192,13 @@ public void resetServiceCheckCount(String checkName) {
         return ServiceCheck.Status.UNKNOWN;
     }
 
+    protected int statusToInteger(String status) {
+        ServiceCheck sc = ServiceCheck.builder()
+                .withStatus(this.statusToServiceCheckStatus(status))
+                .build();
+        return sc.getStatus();
```
AlexandreYang, line 199: Not great way to get the status at integer, but currently the value is private.

PR #298, opened 2020-05-25
Add JMX/RMI socket factory with timeout

prognant:
Proper RMI/JMX socket factory with timeout for both SSL and non-SSL.
Should be a drop in replacement for actual code. I also suggest to remove `Connection.connectWithTimeout` because timeouts are enforced within the socket factory. I also tested the attach API and it seems that infinite freeze can not happen in that case.

`rmi_connection_timeout` & `rmi_client_timeout` have been retained to configure timeouts in the socket factory.

See code comment for a question about the mysterious `attribute.remote.x.request.waiting.timeout` property...

A piece of the code path that may block because of network timeout is located [here](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/jdk8u232-ga/src/share/classes/sun/rmi/transport/tcp/TCPChannel.java#l210) it plays with different configurable timeout during the TCP connection sequence, it even overrides for certain steps the `SO_TIMEOUT` from the the socket factory in at least one [place](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/jdk8u232-ga/src/share/classes/sun/rmi/transport/tcp/TCPChannel.java#l279) (Can be set with through the following property `sun.rmi.transport.tcp.handshakeTimeout`). But having the `SO_TIMEOUT` always set at the socket level ensure always having a configurable timeout on socket operation. The the initial connection timeout is managed differently to keep the ability to get a socket from any factory (used for SSL support at the moment). It shall be noticed that the connection timeout can be shortened but there is an upper bound for enforceable value. The upper bound depends on the OS settings (~1 min on average).

Timeout will then be logged as shown below :

```
2020-06-04 11:51:42,501 | INFO  | App | Could not initialize instance: jmx-10.121.132.56-7198: java.util.concurrent.ExecutionException: java.io.IOException: Failed to retrieve RMIServer stub: javax.naming.CommunicationException [Root exception is java.rmi.ConnectIOException: Exception creating connection to: 10.121.132.56; nested exception is:
        java.io.IOException: connect timed out: 10.121.132.56:7198]
```
For information the call stack while connecting, providing information about involved classes : 

```
PlainSocketImpl.socketConnect(InetAddress,int,int)[native method] (/rt.jar/java.net/PlainSocketImpl.class:-1)
AbstractPlainSocketImpl.doConnect(InetAddress,int,int) (/rt.jar/java.net/AbstractPlainSocketImpl.class:350)
AbstractPlainSocketImpl.connectToAddress(InetAddress,int,int) (/rt.jar/java.net/AbstractPlainSocketImpl.class:206)
AbstractPlainSocketImpl.connect(SocketAddress,int) (/rt.jar/java.net/AbstractPlainSocketImpl.class:188)
SocksSocketImpl.connect(SocketAddress,int) (/rt.jar/java.net/SocksSocketImpl.class:392)
Socket.connect(SocketAddress,int) (/rt.jar/java.net/Socket.class:607)
Socket.connect(SocketAddress) (/rt.jar/java.net/Socket.class:556)
Socket.<init>(SocketAddress,SocketAddress,boolean) (/rt.jar/java.net/Socket.class:452)
Socket.<init>(String,int) (/rt.jar/java.net/Socket.class:229)
RMIDirectSocketFactory.createSocket(String,int) (/rt.jar/sun.rmi.transport.proxy/RMIDirectSocketFactory.class:40)
RMIMasterSocketFactory.createSocket(String,int) (/rt.jar/sun.rmi.transport.proxy/RMIMasterSocketFactory.class:148)
JmxfetchRmiClientSocketFactory.createSocket(String,int) (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/util/JmxfetchRmiClientSocketFactory.java:29)
TCPEndpoint.newSocket() (/rt.jar/sun.rmi.transport.tcp/TCPEndpoint.class:617)
TCPChannel.createConnection() (/rt.jar/sun.rmi.transport.tcp/TCPChannel.class:216)
TCPChannel.newConnection() (/rt.jar/sun.rmi.transport.tcp/TCPChannel.class:202)
UnicastRef.newCall(RemoteObject,Operation[],int,long) (/rt.jar/sun.rmi.server/UnicastRef.class:343)
RegistryImpl_Stub.lookup(String) (/rt.jar/sun.rmi.registry/RegistryImpl_Stub.class:116)
RegistryContext.lookup(Name) (/rt.jar/com.sun.jndi.rmi.registry/RegistryContext.class:132)
GenericURLContext.lookup(String) (/rt.jar/com.sun.jndi.toolkit.url/GenericURLContext.class:205)
InitialContext.lookup(String) (/rt.jar/javax.naming/InitialContext.class:417)
RMIConnector.findRMIServerJNDI(String,Map,boolean) (/rt.jar/javax.management.remote.rmi/RMIConnector.class:1955)
RMIConnector.findRMIServer(JMXServiceURL,Map) (/rt.jar/javax.management.remote.rmi/RMIConnector.class:1922)
RMIConnector.connect(Map) (/rt.jar/javax.management.remote.rmi/RMIConnector.class:287)
JMXConnectorFactory.connect(JMXServiceURL,Map) (/rt.jar/javax.management.remote/JMXConnectorFactory.class:270)
Connection.createConnection() (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/Connection.java:63)
RemoteConnection.<init>(Map) (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/RemoteConnection.java:126)
ConnectionFactory.createConnection(Map) (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/ConnectionFactory.java:38)
Instance.getConnection(Map,boolean) (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/Instance.java:371)
Instance.init(boolean) (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/Instance.java:384)
InstanceInitializingTask.call() (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/InstanceInitializingTask.java:15)
InstanceInitializingTask.call() (/Users/pierre.rognant/git/jmxfetch/src/main/java/org/datadog/jmxfetch/InstanceInitializingTask.java:3)
FutureTask.run() (/rt.jar/java.util.concurrent/FutureTask.class:266)
ThreadPoolExecutor.runWorker(ThreadPoolExecutor$Worker) (/rt.jar/java.util.concurrent/ThreadPoolExecutor.class:1149)
ThreadPoolExecutor$Worker.run() (/rt.jar/java.util.concurrent/ThreadPoolExecutor.class:624)
Thread.run() (/rt.jar/java.lang/Thread.class:748)
```

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Connection.java
+++ b/src/main/java/org/datadog/jmxfetch/Connection.java
@@ -31,17 +31,10 @@ import javax.management.remote.JMXServiceURL;
 public class Connection {
     private static final long CONNECTION_TIMEOUT = 10000;
     public static final String CLOSED_CLIENT_CAUSE = "The client has been closed";
-    private static final ThreadFactory daemonThreadFactory = new DaemonThreadFactory();
     private JMXConnector connector;
     protected MBeanServerConnection mbs;
     protected Map<String, Object> env;
     protected JMXServiceURL address;
-    protected long jmxTimeout = 20;
-
-    private static <T extends Throwable> T initCause(T wrapper, Throwable wrapped) {
-        wrapper.initCause(wrapped);
-        return wrapper;
-    }
 
     /** Gets attributes for matching bean name. */
     public MBeanAttributeInfo[] getAttributesForBean(ObjectName beanName)
@@ -68,7 +61,7 @@ public class Connection {
         this.env.put("attribute.remote.x.request.waiting.timeout", CONNECTION_TIMEOUT);
         closeConnector();
         log.info("Connecting to: " + this.address);
-        connector = connectWithTimeout(this.address, this.env);
+        connector = JMXConnectorFactory.connect(this.address, this.env);
         mbs = connector.getMBeanServerConnection();
     }
 
@@ -81,56 +74,6 @@ public class Connection {
             return ((Attribute) attr).getValue();
         }
         return attr;
-    }
-
-    /**
-     * Connect to a MBean Server with a timeout This code comes from this blog post:
-     * https://weblogs.java.net/blog/emcmanus/archive/2007/05/making_a_jmx_co.html.
-     */
-    JMXConnector connectWithTimeout(final JMXServiceURL url, final Map<String, Object> env)
-            throws IOException {
-
-        final BlockingQueue<Object> mailbox = new ArrayBlockingQueue<Object>(1);
-
-        ExecutorService executor = Executors.newSingleThreadExecutor(daemonThreadFactory);
-        executor.submit(
-                new Runnable() {
-                    public void run() {
-                        try {
-                            JMXConnector connector = JMXConnectorFactory.connect(url, env);
-                            if (!mailbox.offer(connector)) {
-                                connector.close();
-                            }
-                        } catch (Throwable t) {
-                            mailbox.offer(t);
-                        }
-                    }
-                });
-        Object result;
-        try {
-            result = mailbox.poll(jmxTimeout, TimeUnit.SECONDS);
-            if (result == null) {
-                if (!mailbox.offer("")) {
-                    result = mailbox.take();
-                }
-            }
-        } catch (InterruptedException e) {
-            throw initCause(new InterruptedIOException(e.getMessage()), e);
-        } finally {
-            executor.shutdown();
-        }
-        if (result == null) {
-            log.warn("Connection timed out: " + url);
-            throw new SocketTimeoutException("Connection timed out: " + url);
-        }
-        if (result instanceof JMXConnector) {
-            return (JMXConnector) result;
-        }
-        try {
-            throw (Throwable) result;
-        } catch (Throwable e) {
-            throw new IOException(e.toString(), e);
-        }
     }
 
     /** Closes the connector. */
@@ -155,13 +98,5 @@ public class Connection {
             return false;
         }
         return true;
-    }
-
-    private static class DaemonThreadFactory implements ThreadFactory {
-        public Thread newThread(Runnable run) {
-            Thread thread = Executors.defaultThreadFactory().newThread(run);
-            thread.setDaemon(true);
-            return thread;
-        }
     }
 }
--- a/src/main/java/org/datadog/jmxfetch/util/JmxfetchRmiClientSocketFactory.java
+++ b/src/main/java/org/datadog/jmxfetch/util/JmxfetchRmiClientSocketFactory.java
@@ -1,0 +1,133 @@
+package org.datadog.jmxfetch.util;
+
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.net.Socket;
+import java.rmi.server.RMIClientSocketFactory;
+import java.rmi.server.RMISocketFactory;
+import javax.rmi.ssl.SslRMIClientSocketFactory;
+
+public class JmxfetchRmiClientSocketFactory implements RMIClientSocketFactory {
+    private final int timeoutMs;
+    private final int connectionTimeoutMs;
+    private final RMIClientSocketFactory factory;
+
+    /**
+     * JmxfetchRmiClientSocketFactory constructor with socket timeout (milliseconds), a socket
+     * connection timeout (millisecondes) and a flag to enable/disable SSL.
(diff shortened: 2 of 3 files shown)


src/main/java/org/datadog/jmxfetch/Connection.java
```diff
@@ -68,7 +61,7 @@ protected void createConnection() throws IOException {
         this.env.put("attribute.remote.x.request.waiting.timeout", CONNECTION_TIMEOUT);
```
prognant, line 61: I tried to find out where this property might be used. On the "official" oracle side I found something similar only in the the (jmx spec)[https://docs.oracle.com/javase/8/docs/technotes/guides/jmx/JMX_1_4_specification.pdf] with a minor variation (`jmx.remore.x.request.waiting.timeout` listed as a non-standard parameter). So far the only implementation I found that seems to honour this parameter is (weblogic)[https://docs.oracle.com/cd/E13222_01/wls/docs103/jmx/accessWLS.html#wp1112969].

src/main/java/org/datadog/jmxfetch/util/JmxfetchRmiClientSocketFactory.java
```diff
@@ -0,0 +1,30 @@
...
+    public JmxfetchRmiClientSocketFactory(int timeoutMs, boolean ssl) {
+        this.timeoutMs = timeoutMs;
+        this.factory = ssl ? new SslRMIClientSocketFactory()
+                : RMISocketFactory.getDefaultSocketFactory();
+    }
+
+    @Override
+    public Socket createSocket(String host, int port) throws IOException {
+        Socket socket = factory.createSocket(host, port);
+        socket.setSoTimeout(timeoutMs);
```
truthbk, line 26: In the past we had two different timeout values:
- `sun.rmi.transport.proxy.connectTimeout` (connect TO)
- `sun.rmi.transport.tcp.responseTimeout` (read TO)

This socket-level, and more explicit `SO_TIMEOUT` is probably fine for all cases, but we may be potentially changing behavior on new connections by dropping the `rmi_connection_timeout`. It does feel like a bit of a redundant option, but I'm wondering if we didn't have to set that for some weird edge-case.

Also, I may be missing something!

prognant: I removed the usage of `sun.rmi.transport.tcp.responseTimeout` because it sets `SO_TIMEOUT` later in the whole thing lifecyle (as per [1](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/jdk8u232-ga/src/share/classes/sun/rmi/transport/tcp/TCPChannel.java#l99) and [2](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/jdk8u232-ga/src/share/classes/sun/rmi/transport/tcp/TCPChannel.java#l291)) and some earlier operation would occur without having the `SO_TIMEOUT` set.

AFAIK the latter option `sun.rmi.transport.proxy.connectTimeout` _should_ not have impact on our base case as far as my test went, we exit the socket factory [here](https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/jdk/src/share/classes/sun/rmi/transport/proxy/RMIMasterSocketFactory.java#L148), before using the property value. I had a quick look in the code base history I did not find where we were using this property...

prognant: @truthbk I did something similar to the mailbox code (:trollface:) thus we are getting back the `rmi_connection_timeout`, however it's fully located in the socket factory so any subsequent connection not going through `RemoteConnection`  will still be subject to the timeout and I think all case are covered (i.e. SSL/non-SSL, initial connection, subsequent connection directly by the RMI subsystem, timeout during socket lifetime).
@sirianni it should bring an effective & generic solution to the issue #279.

PR #303, opened 2020-06-24
Initial ZGC support

yafernandes:
Initial support for [ZGC](https://wiki.openjdk.java.net/display/zgc/Main) metrics.

### DIFF
--- a/src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
+++ b/src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
@@ -186,3 +186,11 @@
       Usage.used:
         alias: jvm.gc.old_gen_size
         metric_type: gauge
+- include:
+    domain: java.lang
+    type: MemoryPool
+    name: Metaspace
+    attribute:
+      Usage.used:
+        alias: jvm.gc.metaspace_size
+        metric_type: gauge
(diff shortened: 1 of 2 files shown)


yafernandes:
@olivielpeau Yes, tested using `OpenJDK Runtime Environment (build 14.0.1+7)` on MacOS.

vaibhavflip:
@yafernandes @jpbempel need your help to understand is the unit of the metric collected here for pause time? milliseconds?

jpbempel:
yes it's millisecond for CollectionTime
but regarding pause time, it depends on JDK version. for example it's pause time only for JDK14 but it includes concurrent time as well in JDK16.
for JDK17 they resolve the problem with separate MBeans: https://bugs.openjdk.java.net/browse/JDK-8265136
like in Shenandoah see #306

vaibhavflip:
thanks @jpbempel for quick revert. we upgraded our service to jdk17, but post that, we are not getting any gc metrics for pause or collection time anymore. we are on datadog agent `version 7` and using `openjdk version 17.0.2-jdk-slim`. please let me know if we need to do add/change something else as well

jpbempel:
See https://github.com/DataDog/jmxfetch/pull/393

vaibhavflip:
i understand @jpbempel . how shall i go about integrating this with my current datadog? is their a document that i may refer too

PR #258, opened 2020-01-10
[log4j] migrate to log4j2

truthbk:
Let's migrate to log4j2, it's probably about time.

currently testing and fixing some minor stuff

PR #305, opened 2020-06-24
[log4j] migrate to log4j2 (#258)

truthbk:
* [log4j] migrate to log4j2

* [travis] fix log4j config file directive

* [log4j2] support STDERR and STDOUT as ways to refer to log location too


src/main/java/org/datadog/jmxfetch/JmxAttribute.java
```diff
@@ -545,14 +546,18 @@ private String getUserAlias(
...
         alias = alias.replace("$domain", domain);
-        try {
-            alias = alias.replace("$value", getJmxValue().toString());
-        } catch (JMException e) {
-            // If we haven't been able to get the value, it wasn't replaced.
-            LOGGER.warn("Unable to replace $value for attribute " + fullAttributeName, e);
-        } catch (IOException e) {
-            // Same as above
-            LOGGER.warn("Unable to replace $value for attribute " + fullAttributeName, e);
+        if (alias.contains("$value")) {
```
prognant, line 549: Nice catch

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -3,47 +3,110 @@
...
+
+            log.info("File Handler set");
         } else {
-            ConsoleAppender consoleAppender = new ConsoleAppender(new PatternLayout(LOGGER_LAYOUT));
-            if (logLocation != null) {
-                consoleAppender.setTarget(logLocation);
+
+            if (logLocation != null
+                    && (ConsoleAppender.Target.SYSTEM_ERR.toString().equals(logLocation)
+                        || SYSTEM_ERR_ALT.equals(logLocation))) {
```
prognant, line 67: Shouldn't we allow `SYSTEM_OUT`/ `SYSTEM_OUT_ALT` along with  `SYSTEM_ERR`/ `SYSTEM_ERR_ALT`?

truthbk: Maybe.... but this is actually the same on `master`, most of these changes came in from a cherry-pick. I had to fix a few things, but this is really the same everywhere. I don't oppose doing some follow-up and maybe verifying/fixing a couple of things in the logger logic, but this should be 👍 and consistent with our logging on JMXFetch on most recent versions.

Thanks for the review @prognant 🙇

PR #306, opened 2020-06-25
Add Shenandoah metrics

jpbempel:
Use Cycles bean for CollectionCount & CollectionTime.
This includes concurrent time and pauses for the full cycle
Following #303


jpbempel:
tested with JVM running Shenandoah, metrics are reported to DD

PR #314, opened 2020-08-05
Issue debug message for invalid types

AlexandreYang:
When using metric_type in  jmxfetch metrics definition, if the value is anything other than `gauge`, `histogram`, `monotonic_count` it will default to “counter” (aka rate).  https://github.com/DataDog/jmxfetch/blob/f39e75a8a3d5c83890d01f6752cea4dc0aef679c/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java#L122

This seems not ideal since multiple values for `metric_type` can lead this same `counter/rate` behaviour including typos like `gauuge`.

The fix: issue warning/deprecation for anything other than `counter` and `rate`.

We keep both counter and rate as valid since they are already in use and documented here: https://docs.datadoghq.com/integrations/java/?tab=host#the-attribute-filter

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
@@ -95,9 +95,9 @@ public abstract class Reporter {
 
             // StatsD doesn't support rate metrics so we need to have our own aggregator to compute
             // rates
-            if ("gauge".equals(metricType) || "histogram".equals(metricType)) {
+            if (metricType.equals("gauge") || metricType.equals("histogram")) {
                 sendMetricPoint(metricType, metricName, currentValue, tags);
-            } else if ("monotonic_count".equals(metricType)) {
+            } else if (metricType.equals("monotonic_count")) {
                 String key = generateId(metric);
                 if (!instanceCountersAggregator.containsKey(key)) {
                     instanceCountersAggregator.put(key,  currentValue.longValue());
@@ -121,8 +121,8 @@ public abstract class Reporter {
 
             } else {
                 // `counter` and `rate` are equivalent and both accepted as valid.
-                // We continue accepting other cases for backward compatibility.
-                if (!"counter".equals(metricType) && !"rate".equals(metricType)) {
+                // Other types, deprecated or wrong, will default to a counter/rate type and a warning will be logged.
+                if (!metricType.equals("counter") && !metricType.equals("rate")) {
                     log.warn("Invalid metric type " + metricType + " for metric " + metricName
                             + ". The metric will be processed as rate"
                             + " (this is a DEPRECATED behaviour, use a valid type instead).");
--- a/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
@@ -124,7 +124,7 @@ public abstract class Reporter {
                 // Other types, deprecated or wrong, will default to a counter/rate
                 // type and a warning will be logged.
                 if (!metricType.equals("counter") && !metricType.equals("rate")) {
-                    log.warn("Invalid metric type " + metricType + " for metric " + metricName
+                    log.debug("Invalid metric type " + metricType + " for metric " + metricName
                             + ". The metric will be processed as rate"
                             + " (this is a DEPRECATED behaviour, use a valid type instead).");
                 }
--- a/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
@@ -119,7 +119,15 @@ public abstract class Reporter {
                 }
                 sendMetricPoint(metricType, metricName, delta, tags);
 
-            } else { // The metric should be 'counter'
+            } else {
+                // `counter` and `rate` are equivalent and both accepted as valid.
+                // We continue accepting other cases for backward compatibility.
+                if (!"counter".equals(metricType) && !"rate".equals(metricType)) {
+                    log.warn("Invalid metric type " + metricType + " for metric " + metricName
+                            + ". The metric will be processed as rate"
+                            + " (this is a DEPRECATED behaviour, use a valid type instead).");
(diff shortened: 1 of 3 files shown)


AlexandreYang:
Related PR: https://github.com/DataDog/integrations-core/pull/7275

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -119,7 +119,15 @@ public void sendMetrics(List<Metric> metrics, String instanceName, boolean canon
...
 
-            } else { // The metric should be 'counter'
+            } else {
+                // `counter` and `rate` are equivalent and both accepted as valid.
+                // We continue accepting other cases for backward compatibility.
+                if (!"counter".equals(metricType) && !"rate".equals(metricType)) {
+                    log.warn("Invalid metric type " + metricType + " for metric " + metricName
+                            + ". The metric will be processed as rate"
+                            + " (this is a DEPRECATED behaviour, use a valid type instead).");
+                }
```
AlexandreYang, line 129: Note: I tried to have a logic like

```
            if ("gauge".equals(metricType) || "histogram".equals(metricType)) {
                ...
            } else if ("monotonic_count".equals(metricType)) {
                ...
            } else if ("counter".equals(metricType) || "rate".equals(metricType)) {
                sendCounter(...)
            } else {
                // issue warning and deprecation
               sendCounter(...)
            }
```

But the issue is that we will need to pass about 7 args to sendCounter, which is not great. That's why I went to this low changelog approach you see here that only add the warning.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -119,7 +119,15 @@ public void sendMetrics(List<Metric> metrics, String instanceName, boolean canon
                 }
                 sendMetricPoint(metricType, metricName, delta, tags);
 
-            } else { // The metric should be 'counter'
+            } else {
+                // `counter` and `rate` are equivalent and both accepted as valid.
+                // We continue accepting other cases for backward compatibility.
+                if (!"counter".equals(metricType) && !"rate".equals(metricType)) {
```
truthbk, line 125: IMO: this reads better as:
```
 if (!metricType.equals("counter") && !metricType.equals("rate")) {
    ...
}
```
But I just noticed we use this convention elsewhere in the file, so please feel free to ignore or fix both.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -119,7 +119,15 @@ public void sendMetrics(List<Metric> metrics, String instanceName, boolean canon
                 }
                 sendMetricPoint(metricType, metricName, delta, tags);
 
-            } else { // The metric should be 'counter'
+            } else {
+                // `counter` and `rate` are equivalent and both accepted as valid.
+                // We continue accepting other cases for backward compatibility.
```
truthbk, line 124: I would rephrase this:
```
Other types, deprecated or wrong, will default to a counter/rate type and a warning will be logged.
```

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
```diff
@@ -119,7 +119,15 @@ public void sendMetrics(List<Metric> metrics, String instanceName, boolean canon
...
+                if (!"counter".equals(metricType) && !"rate".equals(metricType)) {
+                    log.warn("Invalid metric type " + metricType + " for metric " + metricName
```
truthbk, line 126: I think the warning level is way too verbose, if someone triggers this code path this will be all over the logs.

AlexandreYang: That's right, the upside is that it will motivate to fix the issue.

If we use info, the quantity of output will be same and less visible.

If we use debug, people won't see it and/or ignore it.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
AlexandreYang: Changed for this case and elsewhere.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
AlexandreYang: Changed.

src/main/java/org/datadog/jmxfetch/reporter/Reporter.java
AlexandreYang: @truthbk An alternative is the have an boolean attribute like `this.isInvalidMetricTypeWarningLogged` to log only once:

```java
                if (!metricType.equals("counter")
                        && !metricType.equals("rate")
                        && !this.isInvalidMetricTypeWarningLogged) {
                    log.warn("Invalid metric type " + metricType + " for metric " + metricName
                            + ". The metric will be processed as rate"
                            + " (this is a DEPRECATED behaviour, use a valid type instead).");
                    this.isInvalidMetricTypeWarningLogged = true;
                }
```

WDYT?

truthbk: >  it will motivate to fix the issue
Yeah, that's for sure! 😁 

But some customers ingest agent logs, and those are billed, so I'm pretty sure we shouldn't log at the warn level. As to adding a flag like that, it could work. Printing it just once might not be enough, especially with the log rotation. Maybe lowering it to `debug` is enough, or if you want to make it stand out in `warn`, you could maybe use an integer and set a desired frequency to skip, and print every N iterations or something - it would add some overhead, but only for the bad types so maybe not the end of the world.

AlexandreYang: Let's use debug for now then :)

Changed.

AlexandreYang:
@truthbk Could you have a look again at the PR. The comments have been addressed.

truthbk (approved):

froque:
I have been bitten by the debug level. A warning level should have been used at least once.

PR #321, opened 2020-09-02
Replace jackson databind with lightweight jackson-jr-objects

richardstartin:
Replaces jackson databind (full blown object mapping) with a lightweight version which actually uses the same internals for the JSON mapping use cases (maps, lists, strings, primitives) found in this project, will save 1.3MB in deployment size, builds on #319, #320

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Status.java
+++ b/src/main/java/org/datadog/jmxfetch/Status.java
@@ -1,12 +1,12 @@
 package org.datadog.jmxfetch;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.jr.ob.JSON;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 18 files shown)


src/main/java/org/datadog/jmxfetch/Status.java
```diff
@@ -122,11 +120,11 @@ private String generateYaml() {
         return yaml.dump(status);
     }
 
-    private String generateJson() throws JsonProcessingException {
+    private String generateJson() throws IOException {
```
truthbk, line 123: note: this looks OK to me, but I'm not sure this is well covered in our tests.

truthbk (approved):

src/main/java/org/datadog/jmxfetch/Status.java
richardstartin: OK I will add some testing before merging this

richardstartin: Actually I can see why there isn't much testing for this now - I don't really have time to refactor this for testability right now. Is there another way I can find if this still works the way it's supposed to?

PR #325, opened 2020-09-11
replace guava functionality with equivalent JDK or custom features 

truthbk:
This allows a large (2.7MB) dependency to be dropped from size-sensitive deployments.
Fixes legacy java 1.6 support which broke with previous guava version bump.


olivielpeau (approved):
LGTM overall (haven't reviewed all the code changes in details though).

The similar changes on `master` haven't been QA'ed with the Agent yet, so let's be careful when QA'ing this with Agent v5.

PR #326, opened 2020-09-11
[travis] attempting to add openjdk6 testing

truthbk:
- Adds Java6 to CI
- Fixes broken java6 code


truthbk:
Will not be merged as this is a bit too large a changeset for `agent-v5`, we have merged #327 instead which may not be as thorough but has less entropy.

PR #327, opened 2020-09-14
[travis] adding Java6 testing

truthbk:
This PR is similar to #326, but less aggressive in that it will not remove offending deps/bytecode because the codepath is not susceptible of execution. 

We still Java6 in the CI, so the failing test (which does execute the bad code) will be removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ dist: precise
 
 language: java
 
+install: echo "do nada"
+
 addons:
   hostname: dd-jmxfetch-testhost
   apt:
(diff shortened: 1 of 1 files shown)


olivielpeau (approved):
Thanks, LGTM!

Indeed, the tests you removed execute code paths that should not be called when JMXFetch is used with Agent v5 👍  (there's no strong guarantee that this will remain the case in the future, but I think it's good enough; and in any case it's good to have Java 6 in the CI for the `agent-v5` branch)

PR #330, opened 2020-09-21
added runtime info to stats

gh123man:


truthbk (approved):

src/main/java/org/datadog/jmxfetch/Status.java
```diff
@@ -105,6 +105,7 @@ private void addStats(
         }
         instStats.put("message", message);
         instStats.put("status", status);
+        instStats.put("runtime_version", System.getProperty("java.version"));
```
truthbk, line 108: This is fine, but I think it would make more sense to add the `runtime_version` at the root level, or within an `info` key at the root level. I'm saying this because all JMXFetch instances will be running on the same runtime. So here: https://github.com/DataDog/jmxfetch/blob/a86f53dbd56833096b43241fdb4da6a2a4f180cf/src/main/java/org/datadog/jmxfetch/Status.java#L127 (and for consistency something similar on the `generateYaml()` function too).

That might mean actually having to make changes on the status page side of things which is indeed annoying.

I was thinking something like:

```
========
JMXFetch
========

  Information
  ============
   version: 0.32.8 
   runtime_version: 1.8.0_265

  Initialized checks
  ==================
    jmx
      instance_name : jmx
      message : Unable to instantiate or initialize instance null:9012. Is the target JMX Server or JVM running? Failed to retrieve RMIServer stub: javax.naming.ConfigurationException [Root exception is java.rmi.UnknownHostException: Unknown host: null; nested exception is:
	java.net.UnknownHostException: null]
      metric_count : 0
      service_check_count : 0
      status : ERROR

....
```

I'm trying to think of other stuff we could put there.

truthbk (commented):
Just nitpicking, I'm wondering if a legit info section could make sense. It might be more readable for support purposes.

PR #334, opened 2020-10-13
Throttle bean attribute refresh + Configuration

gh123man:
A new config option to allow bean attribute refresh to be throttled over a time interval, non blocking on a background thread.


djmitche:
@gh123man is this worth rebasing?

gh123man:
This is stale and has been de-prioritized. 
Will re-open and rebase if needed later on.

PR #336, opened 2020-10-30
[status] track dogstatsd client errors

truthbk:
The goal of this PR is to allow us to tap into potential errors happening in the java dogstatsd client that we might otherwise not have good visibility into. This isn't particularly rich, but at least it'll allow support, and us, to quickly identify potential issues on socket operations. This is particularly interesting in the case of UDS, where we could have drops happening at the socket level - this would result in gaps in our customers timeseries - but not be able to prove otherwise.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
@@ -7,8 +7,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.datadog.jmxfetch.Instance;
 import org.datadog.jmxfetch.JmxAttribute;
 
-import java.util.concurrent.atomic.AtomicInteger;
-
 
 /** A reporter class to submit metrics via statsd. */
 @Slf4j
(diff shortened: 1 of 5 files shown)


src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -3,11 +3,13 @@
 import com.timgroup.statsd.NonBlockingStatsDClientBuilder;
 import com.timgroup.statsd.ServiceCheck;
 import com.timgroup.statsd.StatsDClient;
-import com.timgroup.statsd.StatsDClientErrorHandler;
 import lombok.extern.slf4j.Slf4j;
 import org.datadog.jmxfetch.Instance;
 import org.datadog.jmxfetch.JmxAttribute;
 
+import java.util.concurrent.atomic.AtomicInteger;
```
gh123man, line 10: NIT: I don't this import is used in this context

gh123man (approved):

src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
truthbk: Thanks, bad cleanup from my end! :)

gh123man (dismissed):

PR #338, opened 2020-11-10
[MetadataHelper] Close stream opened to read app properties

olivielpeau:
In order to fix a memory leak.

The memory leak only started to be visible with the changes in https://github.com/DataDog/jmxfetch/pull/330 because it started calling this version logic regularly to report JMXFetch's version to the Agent, whereas this logic was previously only called once in a JMXFetch's process lifecycle (either for the `version` command, or to log the version once at startup)

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/util/MetadataHelper.java
+++ b/src/main/java/org/datadog/jmxfetch/util/MetadataHelper.java
@@ -1,17 +1,25 @@
 package org.datadog.jmxfetch.util;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Properties;
 
 public class MetadataHelper {
     /**  Returns our own version number. */
     public static String getVersion() {
+        final Properties properties = new Properties();
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        InputStream stream = classLoader.getResourceAsStream("project.properties");
         try {
-            final Properties properties = new Properties();
-            ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
-            properties.load(classLoader.getResourceAsStream("project.properties"));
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jmxfetch/util/MetadataHelper.java
```diff
@@ -1,19 +1,27 @@
...
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Properties;
 
 public class MetadataHelper {
     /**  Returns our own version number. */
     public static String getVersion() {
+        final Properties properties = new Properties();
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        InputStream stream = classLoader.getResourceAsStream("project.properties");
```
truthbk, line 12: I think this should be enough:

```
try (InputStream stream = classLoader.getResourceAsStream("project.properties")) {
     properties.load(stream);
     return properties.getProperty("version");
} 
```

truthbk (commented):
We should definitely close that input stream, agreed. Nice catch. 

I'm not sure why we're not leaking file descriptors though.... Nor why I saw threads leaking.

src/main/java/org/datadog/jmxfetch/util/MetadataHelper.java
olivielpeau: much better, thanks!

This requires Java 1.7, so I'll change the `maven.compiler` properties accordingly, we only support Java 1.7 and up in the JMXFetch releases shipped with v7/v6 agents anyway.

truthbk (approved):
There's at least one more problem coming from a thread leak in java-dogstatsd-client. That will be addressed in a separate PR there.

Thanks! 🙇

PR #346, opened 2020-12-14
Default cassandra_aliasing to true for cassandra checks

ian28223:
**Description**

- Sets `cassandra_aliasing` to `true` by default for cassandra integration.

**Motivation**

- the config can be easily overlooked if being set via annotations/autodiscovery

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -192,8 +192,12 @@ public class Instance {
         // Alternative aliasing for CASSANDRA-4009 metrics
         // More information: https://issues.apache.org/jira/browse/CASSANDRA-4009
         this.cassandraAliasing = (Boolean) instanceMap.get("cassandra_aliasing");
-        if (this.cassandraAliasing == null) {
-            this.cassandraAliasing = false;
+        if ( this.cassandraAliasing == null) {
+            if ( this.checkName == "cassandra") {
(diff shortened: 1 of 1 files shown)


truthbk (approved):

olivielpeau:
I believe this is a breaking change for users of the cassandra JMXFetch integration, what's the plan to make users aware of the change?

Also this comment needs to be updated: https://github.com/DataDog/jmxfetch/blob/febbaef784d7248db452f6df25b7c747aa5205c5/src/main/java/org/datadog/jmxfetch/JmxAttribute.java#L520-L521

ian28223:
> I believe this is a breaking change for users of the cassandra JMXFetch integration, what's the plan to make users aware of the change?
> 
> Also this comment needs to be updated:
> 
> https://github.com/DataDog/jmxfetch/blob/febbaef784d7248db452f6df25b7c747aa5205c5/src/main/java/org/datadog/jmxfetch/JmxAttribute.java#L520-L521

Cassandra integration users that used the example [cassandra config](https://github.com/DataDog/integrations-core/blob/7.25.x/cassandra/datadog_checks/cassandra/data/conf.yaml.example#L66-L70) should already have this set to `true` by default since before Agent 6.0. The problem is when users manually create the config (e.g. in K8s annotations) and forget to set `cassandra_aliasing=true`, they get metric names that do not match any of the documented [cassandra metrics](https://docs.datadoghq.com/integrations/cassandra/?tab=host#metrics)

PR #348, opened 2020-12-15
Fix empty include in configuration

hush-hush:
An empty 'include' configuration would produce a beanScrope of '*:*' and
would match every bean from the server without applying any of the
default configuration


truthbk (approved):
Looks good to me! If we could add a small unit test to cover this fix, that would be great.

truthbk (approved):
This looks good. 👍 

We could maybe use `getResource()` to avoid using paths the way we do, but I know that pattern was already in the unit tests, so feel free to ignore.

Issue #292, opened 2019-02-01
JVM metrics not showing up until 10 minutes after service starts up


brentm5:
We have have been trying to get custom metrics from resin to show up with the new jmxfetch feature in the APM agent.  We are able to get jvm metrics immediately however our resin metrics do not show up for ~ 10 minutes.

It appears that when the jmxfetch thread starts the resin process has not fully initialized so the resin beans are not available.  After 10 minutes the beans are refreshed and new ones added.  It appears we can get slightly around this by changing the interval to be shorter with the following options however this seems less than ideal. 

```
-Ddd.jmxfetch.refresh-beans-period=30
```

Is there anyway to delay the initial start of the jmx fetch process or double check in the beginning.  I am concerned that setting the beans refresh period to run more frequently could have performance implications.

essobedo:
Fixed by https://github.com/DataDog/jmxfetch/pull/349 and https://github.com/DataDog/dd-trace-java/pull/2659

olivielpeau:
Thanks @essobedo for updating this issue, this is indeed configurable in the Datadog Agent since v7.26.0/v6.26.0 (with the option `refresh_beans_initial` option documented [here](https://docs.datadoghq.com/integrations/java/?tab=host#configuration-options), and should be available in the Java APM client in v0.79.0.

Therefore I'll go ahead and close this issue.

PR #349, opened 2021-01-22
Adds a configurable period for the 1st bean refresh

ian28223:
### What this PR does

Adds a configurable period for the 1st bean refresh.

### Additional info

- Configurable via `refresh_beans_initial` - time in seconds for jmxfetch to wait to refresh the beans for the first time.
- Succeeding bean refresh still adheres to `refresh_beans` parameter.
- Setting a value greater than the `refresh_beans` will default to a value equal to `refresh_beans`.

### Motivation

- AC-737
- Useful for Java applications that are lazy loaded and may take some time after application startup before actually being exposed.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -491,10 +491,10 @@ public class Instance {
 
     /** Returns whather or not the given period has elapsed since reference time. */
     public boolean isPeriodDue(long refTime, Integer refPeriod) {
-        if ((System.currentTimeMillis() - refTime) / 1000 > refPeriod) {
-            return true;
-        } else {
+        if ((System.currentTimeMillis() - refTime) / 1000 < refPeriod) {
             return false;
+        } else {
+            return true;
         }
     }
 
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -441,8 +441,7 @@ public class Instance {
 
         if (this.initialRefreshBeansPeriod != null
                 && (this.initialRefreshTime != 0)
-                && (System.currentTimeMillis() - this.initialRefreshTime) / 1000
-                        > this.initialRefreshBeansPeriod) {
+                && isPeriodDue(this.initialRefreshTime, this.initialRefreshBeansPeriod)) {
             log.info("Refreshing bean list - Initial");
             // We can force the first bean refresh post initialization earlier than the refreshBeansPeriod
             // To enable this, a "refresh_beans_initial" parameter must be specified in the yaml/json config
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -60,6 +60,7 @@
     private static final List<String> MULTI_TYPES =
             Arrays.asList("javax.management.openmbean.TabularData");
     private static final int MAX_RETURNED_METRICS = 350;
+    private static final int DEFAULT_INITIAL_REFRESH_BEANS_PERIOD = 150;
```
truthbk, line 63: IMO the `DEFAULT_INITIAL_REFRESH_BEANS_PERIOD` should be equal to `DEFAULT_REFRESH_BEANS_PERIOD `.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -157,6 +175,7 @@ public Instance(
                 emptyDefaultHostnameObj != null ? (Boolean) emptyDefaultHostnameObj : false;
 
         this.lastCollectionTime = 0;
+        this.initialRefreshTime = 0;
```
truthbk, line 178: I don't think you need this at all, to be honest.

I would rely on `lastRefreshTime` and would do something like the following pseudocode:  
```
period = (this.lastRefreshTime==0) ?  this.initialRefreshBeansPeriod : this.refreshBeansPeriod;

if isPeriodDue(this.lastRefreshTime, period )) {
        // refresh beans
}
```

And as mentioned above `initialRefreshBeansPeriod` should default to `DEFAULT_REFRESH_BEANS_PERIOD`, so that if nothing is set there's no change in behavior.

truthbk (commented):
Just a couple of comments, also, I'm not sure if we currently have a unit test to cover the refresh beans logic, but it would be nice to have otherwise.

src/main/java/org/datadog/jmxfetch/Instance.java
ian28223: Per checking, `lastRefreshTime` is no longer 0 when it gets to this point since it has been updated in [refreshBeansList()](https://github.com/DataDog/jmxfetch/blob/master/src/main/java/org/datadog/jmxfetch/Instance.java#L401) called in init() to initially populate the beans list. Here is the 2nd time it's been called for the 1st _refresh_.
Still, I've updated the code inspired from this. Thanks!

ian28223:
> Just a couple of comments, also, I'm not sure if we currently have a unit test to cover the refresh beans logic, but it would be nice to have otherwise.
> 

I'd work on the unit test either here (if it doesn't make it before freeze).. or in another PR

truthbk (approved):

tylerbenson:
I think this resolves #292.

PR #351, opened 2021-01-29
Add tests for refresh_beans and refresh_beans_initial

ian28223:
Related to #349 

Adds tests for both `refresh_beans_initial` and `refresh_beans`

- AC-800

### DIFF
--- a/src/test/java/org/datadog/jmxfetch/TestInstance.java
+++ b/src/test/java/org/datadog/jmxfetch/TestInstance.java
@@ -147,4 +147,52 @@ public class TestInstance extends TestCommon {
 
         assertEquals(2, configurationList.size());
     }
+
+    /** Tests refresh_beans_initial and the following refresh_beans */
+    @Test
+    public void testRefreshBeans() throws Exception {
+        SimpleTestJavaApp testApp = new SimpleTestJavaApp();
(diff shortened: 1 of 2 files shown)


sgnn7 (dismissed):
@ian28223 LGTM. I do feel like sleeping in unit tests is a bit of a problematic pattern but  I don't know if there's a better way to do this (without reworking the code heavily).

ian28223:
> @ian28223 LGTM. I do feel like sleeping in unit tests is a bit of a problematic pattern but I don't know if there's a better way to do this (without reworking the code heavily).

I've reduced the sleep time to cut down on the overall runtime of the test. I hope this helps a bit

sgnn7:
@ian28223 This LGTM but we should probably wait for @truthbk to maybe add some feedback if we can remove the `sleep`s entirely.

sgnn7 (approved):

truthbk (commented):
The issue with the sleeps is not as much a problem with test duration (though it's always nice to keep these as short as possible), but rather with the fact it often turns out to be a bit of an  anti-pattern that introduces flaky tests.

It's unclear to me, with the current code, how we can avoid sleeping at all, so I think I'm fine with this.

PR #358, opened 2021-04-13
Add `jvm.gc.old_gen_size` as an alias for `Tenured Gen`

ogaca-dd:
Add `jvm.gc.old_gen_size` as an alias for `Tenured Gen`.

Motivation:
`SerialGC` may be the default GC for `openjdk 11.0.10` and it exposes `Tenured Gen` beans.

See AC-907.

### DIFF
--- a/src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
+++ b/src/main/resources/org/datadog/jmxfetch/default-jmx-metrics.yaml
@@ -189,6 +189,14 @@
 - include:
     domain: java.lang
     type: MemoryPool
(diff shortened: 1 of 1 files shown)


olivielpeau (approved):
LGTM if you've confirmed this works well!

About the name of this new metric: I was wondering if `jvm.gc.old_gen_size` was the right name or if we should use `jvm.gc.tenured_gen_size`, but apparently "tenured" and "old" are synonyms (according to https://stackoverflow.com/a/26462943 anyway), so using `jvm.gc.old_gen_size` makes sense for consistency with the other metrics 👍

PR #368, opened 2021-07-09
Move CI to CircleCI

vickenty:


djmitche:
Does this replace #217 and #263?

PR #369, opened 2021-08-16
Remove messages for non-ERROR can_connect service checks

yzhan289:
This PR removes messages for status that is not an ERROR for the `*.can_connect` service check.

Datadog is removing messages for integration checks that give OK status to reduce the load on the backend. There's a small scenario where `*.can_connect` gives a WARNING status along with a message. This WARNING is later converted into an OK, as only OK or ERROR statuses are used for `*.can_connect`. 

Once this is merged in, https://github.com/DataDog/integrations-core/pull/9958 can be merged.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -1191,7 +1191,6 @@ public class App {
                                     + instance.getMaxNumberOfMetrics()
                                     + " metrics.";
 
-                    instanceStatus = Status.STATUS_OK;
                     CustomLogger.laconic(log, Level.WARN, instanceMessage, 0);
                 }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -839,6 +839,11 @@ private void sendServiceCheck(
     private void sendCanConnectServiceCheck(Reporter reporter, String checkName,
         String serviceCheckPrefix, String status, String message, String[] tags) {
         String serviceCheckName = String.format("%s.can_connect", serviceCheckPrefix);
+
+        if (status != Status.STATUS_ERROR) {
```
ofek, line 843:
```suggestion
        if (status == Status.STATUS_OK) {
```

ofek (changes requested):

src/main/java/org/datadog/jmxfetch/App.java
ofek: then remove https://github.com/DataDog/jmxfetch/blob/f398b502c30f409401bfdb21e57339119dcc1ba1/src/main/java/org/datadog/jmxfetch/App.java#L1194

ofek (approved):

remeh (approved):
LGT

PR #374, opened 2021-11-03
[services] adding multi-service metric forking facilities

truthbk:
This PR adds support for multiple service reporting. This has important implications as the *same* metric will be reported multiple times, tagged by each service. You will get N time-series, one for each service defined. 

While the above will work well for those time-series as individually scoped, it will also have important implications in multiple aggregation scenarios. While the `avg` will typically be fine in-app, and correlation for APM *will* work, this can also give place to weird/confusing queries, where a `sum` of these time-series could give place to outright wrong values (say a `sum:jvm.heap.*` would be added N times, once per service, yet all those services are running on the same JVM).

An alternative to this would be to submit the same time-series tagged by multiple time-series. That would probably address the dashboard issues, but could result in correlation problems for services (if time-series are assumed to be tagged with a single service tag).


truthbk:
Closed in favor of #375

PR #375, opened 2021-11-09
[services] support adding multiple service tags to submitted metrics

truthbk:
This PR is an alternative approach to #374. This proposal, instead of submitting multiple "copies" of the same instance with different `service` tags as we implemented in the aforementioned PR, will send a single time-series tagged with multiple service tags. 

The approach proposed here should allow for more sensible querying as the datadog backend is able to identify this consists of a *single* time-series despite the various `service` tags and thus provides more sensible values when dealing with aggregation functions like `sum`. 

Before deciding between this approach or #374 we need to understand if what we do here may cause issues with service correlation. 

### Implementation details

This introduces a `ServiceNameProvider` interface, an implementation of which can be passed to JMXFetch's `AppConfig` builder:
* In a standalone JMXFetch scenario, a default implementation of this provider (`ConfigServiceNameProvider`) is used, which pulls the `service` defined in yaml config. This implementation supports defining multiple services under `service` in configuration.
* When configured on `AppConfig` (typically from the APM tracer):
  
  * the configured `ServiceNameProvider` applies to all the JMXFetch Instances that are configured on JMXFetch (i.e.: when a `ServiceNameProvider` is configured on JMXFetch, all the metrics submitted by JMXFetch will be submitted with the service names provided by `ServiceNameProvider`). No possibility at the moment to configure this at the level of a single JMXFetch `Instance`.
  * the configured `ServiceNameProvider` does not read the `service` defined in yaml config. Note that for the APM tracer, this can constitute a breaking change for APM users that have defined custom JMXFetch yaml configs in which they've defined a `service`: previously JMXFetch would send both the tracer-provided `service` name and the config-provided `service`, with this change JMXFetch will only send the trace-provided `service`.

* it's best for the `ServiceNameProvider` to return an immutable List, to avoid race conditions/thread-safety issues.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -581,7 +581,7 @@ public class Instance {
                             && !action.equals(AppConfig.ACTION_LIST_NOT_MATCHING)) {
                         reporter.displayMetricReached();
                         metricReachedDisplayed = true;
-                            }
+                    }
                 }
                 JmxAttribute jmxAttribute;
                 String attributeType = attributeInfo.getType();
@@ -674,7 +674,7 @@ public class Instance {
                                     && limitReached) {
                                 reporter.displayMatchingAttributeName(
                                         jmxAttribute, metricsCount, maxReturnedMetrics);
-                                    }
+                            }
                             break;
                         }
                     } catch (Exception e) {
@@ -691,7 +691,7 @@ public class Instance {
                         && (action.equals(AppConfig.ACTION_LIST_EVERYTHING)
                             || action.equals(AppConfig.ACTION_LIST_NOT_MATCHING))) {
                     reporter.displayNonMatchingAttributeName(jmxAttribute);
-                            }
+                }
             }
         }
         log.info("Found " + matchingAttributes.size() + " matching attributes");
@@ -750,8 +750,8 @@ public class Instance {
             }
         }
 
-        if(this.services != null) {
-            for(String service : this.services){
+        if (this.services != null) {
+            for (String service : this.services) {
                 tags.add("service:" + service);
             }
         }
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -87,7 +87,7 @@ public class Instance {
     private Map<String, Object> instanceMap;
     private Map<String, Object> initConfig;
     private String instanceName;
-    private String service;
+    private List<String> services;
     private Map<String, String> tags;
     private String checkName;
     private String serviceCheckPrefix;
@@ -157,12 +157,9 @@ public class Instance {
             this.initialRefreshBeansPeriod = this.refreshBeansPeriod;
         }
 
-        this.service = (String) instanceMap.get("service");
-        if ((this.service == null || this.service.isEmpty()) && initConfig != null) {
-            this.service = (String) initConfig.get("service");
-        }
-        if (this.service != null && !this.service.isEmpty()) {
-            this.tags.put("service", this.service);
+        this.services = compileServiceList(instanceMap);
+         if (services.size() == 0) {
+            services = compileServiceList(initConfig);
         }
 
         this.minCollectionPeriod = (Integer) instanceMap.get("min_collection_interval");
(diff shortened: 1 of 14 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -256,6 +254,32 @@ public Instance(
         }
     }
 
+    /** this function is unsafe, please call holding a lock. */
+    public synchronized void updateServiceList(List<String> services) {
+        this.services = services;
```
devinsba, line 259: I might add a null check here, not that we would try to hand a null list but it's probably best to ensure a list with some default service name at the least

devinsba: Possibly even make an unmodifiable copy here too

devinsba (commented):
I don't think the instance methods need to be synchronized. I don't see any concurrent modification type exceptions because the list is being replaced instead of modified.

src/main/java/org/datadog/jmxfetch/Instance.java
truthbk: @devinsba the service is not a guaranteed tag in the jmxfetch world, so we have to be ready for a null list. The rest of the code should be resilient to handling `Instance.services` being null.

Finally, please correct me if I'm wrong - replacing a reference variable is not an atomic operation, so if we're going to be accessing `this.services` concurrently wouldn't we need to have some synchronization? If we don't I'm happy to remove the synchronization, but let's keep in mind the scenario where we call this function somehow from the tracer at some random time to update the service list.

devinsba: Yes. Reference writes are atomic https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html

> Reads and writes are atomic for reference variables and for most primitive variables (all types except long and double).

In this case the synchronization wouldn't accomplish the goal I don't think since synchronized methods all have their own lock. They don't share

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -256,6 +254,32 @@ public Instance(
         }
     }
 
+    /** this function is unsafe, please call holding a lock. */
+    public synchronized void updateServiceList(List<String> services) {
```
devinsba, line 258: Instead of having a method where we pass the list of services, what do you think about adding a constructor parameter that is an instance of an interface like:

```
interface ServiceNameProvider {
  List<String> getServiceNames();
}
```
Then we can own that object and you can create a simple one for the basic case

olivielpeau:
@devinsba Jaime is OOO at the moment, so I took a stab at changing the API a bit to (hopefully) integrate more easily with the java tracer:

* now uses a ServiceNameProvider interface, an implementation of which can be passed to JMXFetch's `AppConfig` builder
* when configured on AppConfig, the configured ServiceNameProvider applies to all the JMXFetch `Instance`s that are configured on JMXFetch (i.e.: when a ServiceNameProvider is configured on JMXFetch, all the metrics submitted by JMXFetch will be submitted with the service names provided by ServiceNameProvider). No way at the moment to configure this at the level of a single JMXFetch `Instance`
* with the current implementation, I think it's best for the ServiceNameProvider to return an immutable `List`, to avoid race conditions/thread-safety issues

Let me know what you think! Also let me know if you need me to cut a release of JMXFetch so you can start integrating these changes in the java tracer.

devinsba (approved):

devinsba:
I tested this, it works as long as the tracer is configured with an environment (`DD_ENV` or `dd.env`). I'm good with merging and releasing as soon as you all are comfortable

src/main/java/org/datadog/jmxfetch/JmxAttribute.java
```diff
@@ -136,9 +137,12 @@ private void addAdditionalTags() {
...
     }
 
-    private void addServiceTag() {
-        if (serviceName != null && !serviceName.isEmpty()) {
-            this.defaultTagsList.add("service:" + serviceName);
+    private void addServiceTags() {
+        Iterable<String> serviceNames = this.serviceNameProvider.getServiceNames();
+        if (serviceNames != null) {
+            for (String serviceName : serviceNames) {
+                this.defaultTagsList.add("service:" + serviceName);
```
devinsba, line 144: Does order matter with these calls? ie: do we need to return a stably ordered collection?

truthbk: Order does not matter, we should be good.

PR #376, opened 2021-11-09
[app] no need for run to be static; use constructor.

truthbk:
This is a nasty pattern. Instantiating an instance to the app in a static method straight up feels wrong. Promoting a pattern where the caller might not even get a copy to the app object in case there's a need to interact with it feels wrong.

I'd rather have the caller instantiate and then launch the run routine explicitly.  

There's room for a bigger refactor, but this PR should suffice to get rid of the anti-pattern.


sgnn7 (approved):
LGTM!

devinsba (approved):
Makes sense from the agent point of view.

PR #383, opened 2021-12-20
Use `slf4j`+`java.util.logging` for logging needs.

remeh:
JMXFetch needs for logging features are fairly small, this PR is switching all the logging to use `java.util.logging` SLF4J bindings.

All features (file logging with rotation, stdout, stderr, log levels, rfc3339 date time format support) should still be present.

Some rough performances tests have been used to confirm that JMXFetch still performs well when it logs a lot, even if it is not really verbose in default scenarios.

This PR bumps SLF4J from `1.7.26` to `1.7.32`.


src/main/java/org/datadog/jmxfetch/util/LogLevel.java
```diff
@@ -0,0 +1,114 @@
...
+            return DEBUG;
+        } else if (julLevel == Level.FINER) {
+            return TRACE;
+        } else if (julLevel == Level.FINEST) {
+            return TRACE;
+        } else if (julLevel == Level.OFF) {
+            return OFF;
+        }
+        // TODO(remy): should we fallback on OFF?
+        // (jaime): IMHO nope.
```
olivielpeau, line 73: agreed. Let's remove the comments?

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
```diff
@@ -0,0 +1,114 @@
...
+        if (str.toUpperCase().equals("FATAL")) {
+            return ERROR;
+        }
+        for (LogLevel l : LogLevel.class.getEnumConstants()) {
+            if (str.toUpperCase().equals(l.toString().toUpperCase())) {
+                return l;
+            }
+        }
+        // TODO(remy): should we fallback on OFF?
+        // (jaime): IMHO nope.
```
olivielpeau, line 89: agreed, let's remove?

olivielpeau (commented):
Thanks! I haven't reviewed the `CustomLogger` changes in details, but the log level conversion looks good to me. Left a few nit comments.

pom.xml
```diff
@@ -39,8 +39,6 @@
         <apache-commons-lang3.version>3.5</apache-commons-lang3.version>
         <java-dogstatsd-client.version>2.10.5</java-dogstatsd-client.version>
         <jcommander.version>1.35</jcommander.version>
-        <!-- log4j 2.13+ drops support for Java 7, so stick to 2.12 -->
-        <log4j.version>2.12.2</log4j.version>
         <slf4j.version>1.7.26</slf4j.version>
```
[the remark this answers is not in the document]
remeh: Good idea.

.gitignore
```diff
@@ -19,3 +19,4 @@ target/*
 
 # jenv
 .java_version
+.java-version
```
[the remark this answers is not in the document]
truthbk: It's not a duplicate, one is underscrote, the other is a regular hyphen

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.logging.Level;
+
+/**
+ * LogLevel used for internal logging to match Datadog Agent levels.
+ * Comparison table with java.util.logging:
+ * <pre>
+ *  JUL     |  JMXFetch LogLevel
+ * ----------------------------
+ *  ALL     |  ALL
+ *  SEVERE  |  ERROR
```
sgnn7, line 12: Would be good to also mention `FATAL` here

.gitignore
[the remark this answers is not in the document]
olivielpeau: My bad! We can keep this then

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
```diff
@@ -0,0 +1,114 @@
...
+        this.level = level;
+        this.label = label;
+    }
+
+    // --
+
+    /** contains returns if the given log level is contained.
+     * If this level is `OFF`, `contains` returns `true` only if `other` is `OFF`
+     * as well.
+     */
```
sgnn7, line 43: nit: Javadoc-like comments in this file probably need a bit standardization/formatting

.gitignore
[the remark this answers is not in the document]
remeh: I kept it and resolved already ;)

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
```diff
@@ -0,0 +1,114 @@
...
+    public boolean contains(LogLevel other) {
+        if (this == OFF) {
+            return this.level == other.level;
```
sgnn7, line 46: I'm guessing more appropriate version of this line should be:
```suggestion
            return other.level == OFF;
```

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+        return this.level <= other.level;
+    }
+
+    /** fromJulLevel converts a java.util.logging.Level into a LogLevel. */
+    public static LogLevel fromJulLevel(Level julLevel) {
+        if (julLevel == Level.ALL) {
```
sgnn7, line 53: Is there a reason we don't use `switch` here?

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
+    private static final String DATE_JDK14_LAYOUT = "yyyy-MM-dd HH:mm:ss z";
+    private static final String DATE_JDK14_LAYOUT_RFC3339 = "yyyy-MM-dd'T'HH:mm:ssXXX";
+    private static final String JDK14_LAYOUT = "%s | JMX | %2$s | %3$s | %4$s%n";
 
-    /** Sets up the custom logger to the specified level and location. */
-    public static void setup(Level level, String logLocation, boolean logFormatRfc3339) {
-        final LoggerContext ctx = (LoggerContext) LogManager.getContext(false);
-        final Configuration config = ctx.getConfiguration();
+    private static boolean isStdErr(String target) {
+        List<String> stderrs = Arrays.asList("SYSTEM.ERR", "SYSTEM_ERR", "STDERR");
```
sgnn7, line 39: nit: Should the `Arrays.asList` in these functions be `static` themselves? I'm guessing either way it's not a hot path and it doesn't matter so it's not a blocker.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
-                ctx.getRootLogger().addAppender(config.getAppender(ca.getName()));
+            @Override
+            public synchronized String format(LogRecord lr) {
+                SimpleDateFormat simpleDateFormat = new SimpleDateFormat(dateFormat,
+                                                        Locale.getDefault());
+                return String.format(format,
+                    simpleDateFormat.format(new Date()).toString(),
+                    // NOTE(remy): these conversions may generate a lot of garbage
+                    //      (jaime): if performance impact over a sustained period of time
+                    //               isn't too bad, we can probably live with this.
```
sgnn7, line 80: The comment in the code here shouldn't end up getting merged as-is

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
-                    .withLayout(layout)
-                    .build();
-
-                ca.start();
-                config.addAppender(ca);
-                ctx.getRootLogger().addAppender(config.getAppender(ca.getName()));
+            @Override
+            public synchronized String format(LogRecord lr) {
+                SimpleDateFormat simpleDateFormat = new SimpleDateFormat(dateFormat,
+                                                        Locale.getDefault());
```
sgnn7, line 75: Can this variable be pulled out of the `@Override` itself since it's in the hot path?

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
+        if (stderrHandler != null) {
+            logger.addHandler(stderrHandler);
+        }
+    }
+
+    /** closeHandlers closes all opened handlers. */
+    public static synchronized void shutdown() {
+        Logger logger = Logger.getLogger("");
+        for (Handler handler : logger.getHandlers()) {
+            handler.close();
```
sgnn7, line 150: We likely want to remove the handlers first from the logger. We should also flush the handlers before closing I think?

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
+
+        // the logLocation isn't always containing a file, it is sometimes
+        // referring to a standard output. We want to create a FileHandler only
+        // if the logLocation is a file on the FS.
+        if (logLocation != null && logLocation.length() > 0) {
+            if (!isStdOut(logLocation) && !isStdErr(logLocation)) {
+                // file logging
+                try {
+                    // maximum one 5MB file
+                    fileHandler = new FileHandler(logLocation, 5 * 1024 * 1024, 1);
```
sgnn7, line 103: The 5mb value should be pulled up to consts

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
+                    fileHandler.setFormatter(formatter);
+                } catch (Exception e) {
+                    fileHandler = null;
+                    log.error("can't open the file handler");
```
sgnn7, line 107: We should print at least `e.getMessage()` value here for debugging

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
+        // --
+
+        ConsoleHandler stdoutHandler = null;
+        ConsoleHandler stderrHandler = null;
+        FileHandler fileHandler = null;
+
+        // the logLocation isn't always containing a file, it is sometimes
+        // referring to a standard output. We want to create a FileHandler only
+        // if the logLocation is a file on the FS.
+        if (logLocation != null && logLocation.length() > 0) {
```
sgnn7, line 98: nit: The nested if block I think could be decomposed a bit but it's not a big deal and not a blocker.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,166 @@
...
+        // ---
 
-        loggerConfig.addAppender(appender, null, null);
-        loggerConfig.setLevel(level);
+        Logger logger = Logger.getLogger("");
 
-        ctx.updateLoggers();
+        // clean all existing handlers
+        for (Handler handler : logger.getHandlers()) {
+            logger.removeHandler(handler);
```
sgnn7, line 130: I'm guessing we want to flush the handlers before removing. They also may need closing.

sgnn7 (commented):
@remeh Left some comments but it's mostly nits. Great work on this!

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
remeh: definitely was the intended change 🤦 fixing that, good catch 🙇

remeh: we don't really mind this handler, but that's not harming anyone so I'll close them first 👍

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
remeh: yeah this one isn't my best.. I've tried to replicate previous behavior, but with easier to read if blocks. We may want to revisit this

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
remeh: you're right

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
remeh: > The close method will perform a flush and then close the Handler

I'll remove the handler though 👍

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
remeh: `java.util.logging.Level` can't be used in a `switch`, at least not in Java 7 nor 8.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
remeh: removing it

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
remeh: I think you're right

sgnn7 (approved):

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -48,6 +47,7 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.LogManager;
```
olivielpeau, line 50: unused import AFAICT? (I'm surprised the linter doesn't catch this)

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,173 @@
...
 
 @Slf4j
 public class CustomLogger {
     private static final ConcurrentHashMap<String, AtomicInteger> messageCount
             = new ConcurrentHashMap<String, AtomicInteger>();
+
     private static final String LAYOUT = "%d{yyyy-MM-dd HH:mm:ss z} | JMX | %-5p | %c{1} | %m%n";
 
     private static final String LAYOUT_RFC3339 =
         "%d{yyyy-MM-dd'T'HH:mm:ss'Z'} | JMX | %-5p | %c{1} | %m%n";
```
olivielpeau, line 32: these are unused now, let's remove them

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,173 @@
...
+
+        // the logLocation isn't always containing a file, it is sometimes
+        // referring to a standard output. We want to create a FileHandler only
+        // if the logLocation is a file on the FS.
+        if (logLocation != null && logLocation.length() > 0) {
+            if (!isStdOut(logLocation) && !isStdErr(logLocation)) {
+                // file logging
+                try {
+                    // maximum one 5MB file
+                    fileHandler = new FileHandler(logLocation, MAX_FILE_SIZE, FILE_COUNT);
```
olivielpeau, line 104: If I'm reading the javadocs right, I think this does a different thing than we did before: now, with `FILE_COUNT` set to 1, we only write to one file and don't really rotate it (see https://docs.oracle.com/javase/7/docs/api/java/util/logging/FileHandler.html#FileHandler(java.lang.String,%20int,%20int), I guess the logger just writes to the same file again once it's reached the max size (not sure).
Also the `logLocation` that we pass is the absolute path to the log file, but instead this expects a pattern, see the top of the docs page: https://docs.oracle.com/javase/7/docs/api/java/util/logging/FileHandler.html

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,173 @@
...
-                    .withName(target)
-                    .withLayout(layout)
-                    .build();
-        appender.start();
+        // should configure the root logger
+        // ---
 
-        loggerConfig.addAppender(appender, null, null);
-        loggerConfig.setLevel(level);
+        Logger logger = Logger.getLogger("");
```
olivielpeau, line 127: I think we need to keep a reference to this logger (maybe at the level of the `CustomLogger` class) otherwise the logger might get garbage-collected if I read https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#getLogger(java.lang.String) correctly (in which case I think it'd get completely reset after getting GC'ed).

Since `""` returns the root logger, I may have missed some additional documentation that explains that the root logger never gets GC'ed maybe 🤔

src/main/java/org/datadog/jmxfetch/App.java
sgnn7: Interestingly, I ran `checkstyle` linter locally and it also didn't see this line as a problem.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
sgnn7: I think from reading this, it's a bit ambiguous in the docs about what happens here but the behavior might match:

From the intro to the class
```
For a rotating set of files, as each file reaches a given size limit, it is closed, rotated out, and a new file opened. Successively older files are named by adding "0", "1", "2", etc. into the base filename.
```
This makes me think that the code should maintain a single rotating log file like it did previously but I could be wrong (I'm going to test this locally to see what happens)

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
sgnn7: Looks like root logger doesn't need a strong ref afaict: https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/util/logging/LogManager.java#L464-L465

olivielpeau: Good find! Should this be regarded as an implementation detail that may be different in other JVM implementations? (I'd feel safer with a strong ref to the root logger in a class attribute, but I may be a bit paranoid)

sgnn7: I think this isn't an implementation detail since the root logger holds the default handlers. With that said, I don't think there would be harm in using a strong ref just in case.

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
```diff
@@ -0,0 +1,121 @@
...
+ * </p>
+ */
+public enum LogLevel {
+    ALL(0, "ALL"),
+    ERROR(1, "ERROR"),
+    WARN(2, "WARN"),
+    INFO(3, "INFO"),
+    DEBUG(4, "DEBUG"),
+    TRACE(5, "TRACE"),
+    OFF(6, "OFF");
```
olivielpeau, line 33: This order looks a bit weird to me. I would expect ALL to be number 6, and OFF number 0 to follow the order, but I might be missing something. Also I haven't checked where this order is used

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
olivielpeau: Fixed now that you've changed `FILE_COUNT` to `2`

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
truthbk: I don't think we need a strong reference in our code.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,192 @@
...
+        if (stderrHandler != null) {
+            logger.addHandler(stderrHandler);
+        }
+
+        // filter all other log handlers
+        for (Enumeration<String> logNames = manager.getLoggerNames() ;
+                logNames.hasMoreElements() ; ) {
+            String logName = logNames.nextElement();
+
+            for (Handler handler : manager.getLogger(logName).getHandlers()) {
```
olivielpeau, line 144: According to java docs, `manager.getLogger` can return `null` so we need to check for that value  https://docs.oracle.com/javase/7/docs/api/java/util/logging/LogManager.html#getLogger(java.lang.String)

olivielpeau: also, I think we might as well just remove the handlers we find instead of setting a filter on them?

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,192 @@
...
+        // set our jmxfetch handlers
+        for (Handler handler : logger.getHandlers()) {
+            handler.setFilter(new Filter() {
+                @Override
+                public boolean isLoggable(LogRecord record) {
+                    String sourceClass = record.getSourceClassName();
+                    return sourceClass.startsWith("org.datadog.jmxfetch");
+                }
+            });
+        }
```
olivielpeau, line 166: I think we can achieve this by creating our logger named `"org.datadog.jmxfetch"`. Since all the JMXFetch classes use the `@Slf4j` annotation, they use a `log`ger that's a child of `org.datadog.jmxfetch` (https://projectlombok.org/api/lombok/extern/slf4j/Slf4j.html)

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
olivielpeau: actually, I think all of this is unnecessary because `manager.reset();` above should already remove all of the handlers: https://docs.oracle.com/javase/7/docs/api/java/util/logging/LogManager.html#reset():

> For all named loggers, the reset operation removes and closes all Handlers and (except for the root logger) sets the level to null. The root logger's level is set to Level.INFO.

So, once `manager.reset()` is called, I think all existing handlers are removed.

olivielpeau (commented):
I left a few comments that I'll address right away

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -230,8 +226,7 @@ private static void attachShutdownHook() {
                 @Override
                 public void run() {
                     log.info("JMXFetch is closing");
-                    // Properly close log handlers
-                    LogManager.shutdown();
+                    // make sure log handlers are properly closed
```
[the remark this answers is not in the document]
olivielpeau: re-introduced

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
olivielpeau: (openJDK source code, for reference: https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/util/logging/LogManager.java#L832-L845 )

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
olivielpeau: I've addressed this, resolving

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
remeh: Addressed.

src/main/java/org/datadog/jmxfetch/util/LogLevel.java
olivielpeau: `FATAL` shouldn't be in this table because it's not a `LogLevel`, it's only supported as a config log level string that's then converted to `LogLevel.ERROR`, as mentioned below

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
```diff
@@ -1,123 +1,168 @@
...
+            }
+        }
     }
 
     /** Laconic logging for reduced verbosity. */
-    public static void laconic(org.slf4j.Logger logger, Level level, String message, int max) {
+    public static void laconic(org.slf4j.Logger logger, LogLevel level, String message, int max) {
         if (shouldLog(message, max)) {
-            if (level.isInRange(Level.ERROR, Level.ALL)) {
+            if (level == LogLevel.ERROR || level == LogLevel.ALL) {
```
olivielpeau, line 159: We should change this to:

```suggestion
            if (level == LogLevel.ERROR) {
```

There's no reason `laconic` would be called to log a message with a log level of `LogLevel.ALL`. Its only use today it 1 call with `LogLevel.WARN`.

src/main/java/org/datadog/jmxfetch/util/CustomLogger.java
olivielpeau: Yes, since it's not in a hot path I won't change this now, resolving.

olivielpeau (approved):
PR LGTM

olivielpeau (approved):
Re-approving to account for Remy's latest changes.

I've QA'ed the current state of the branch with an agent, LGTM.

PR #392, opened 2022-04-22
Update jcommander to avoid unsafe dependency resolution vulnerability

kylegearhart:
## What's in this PR?
- update to jcommander dependency to avoid: https://security.snyk.io/vuln/SNYK-JAVA-COMBEUST-174815
- jcommander fixed the issue in [this commit](https://github.com/cbeust/jcommander/commit/3ae95595febbed9c13f367b6bda5c0be1c572c53)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
         <commons-io.version>2.4</commons-io.version>
         <apache-commons-lang3.version>3.5</apache-commons-lang3.version>
         <java-dogstatsd-client.version>2.10.5</java-dogstatsd-client.version>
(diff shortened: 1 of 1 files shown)


truthbk (approved):

truthbk:
This was actually my bad @kylegearhart we can't quite merge this as-is because there are breaking changes in the version bump, and, on top of that, it breaks Java7 support. Unfortunately, we still have some customers on very old legacy JVMs and we still try to keep Java7 support. Let me revert this while we consider potential alternatives.

PR #395, opened 2022-06-23
Revert "Update jcommander to avoid unsafe dependency resolution vulnerability"

truthbk:
Reverts DataDog/jmxfetch#392


truthbk:
We know this reverts back to an older version we should bump, but let's discuss our alternatives first.

PR #403, opened 2022-12-07
Prevent agent from adding origin detection tags when using UDS.

vickenty:
dd.internal.entity_id:none, which jmxfetch already uses, only works with non-default agent configuration (dogstatsd_entity_id_precedence: true).

To avoid this issue, agent 7.41.0 supports new dd.internal.card:none which also opts-out from the origin detection, but without the need for non-default configuration. The dd.interna.card tag prefix is recognized since 7.27.0, and will be removed from metrics.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
@@ -49,7 +49,7 @@ public class StatsdReporter extends Reporter {
                 .errorHandler(handler)
                 .entityID(entityId);
 
-        // When using UDS set the datagram size to 8k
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -52,6 +52,7 @@ private void init() {
         // When using UDS set the datagram size to 8k
```
gh123man, line 52:
```suggestion
        // When using UDS set the datagram size to 8k and disable origin detection
```
We could update this comment since it does more than set the size to 8k now.

gh123man (dismissed):

gh123man (approved):

PR #416, opened 2023-03-29
Statsd client blocks by default allowing backpressure instead of dropping data

scottopell:
The default mode for java-dogstatsd-client is to be non-blocking as this is what most java applications want (submitting some telemetry data should not hold up the rest of their application), however for JMXFetch, there is nothing else to be delayed by doing blocking IO.

Blocking IO by default should allow large batches of metrics to block (not drop) if the underlying socket is temporarily full, and since JMXFetch only collects data on a relatively large interval (15s by default), there should be plenty of time for any back-pressure to be drained.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
+++ b/src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
@@ -18,13 +18,15 @@ public class StatsdReporter extends Reporter {
     private Boolean telemetry;
     private int queueSize;
     private long initializationTime;
+    private boolean nonBlocking;
 
     /** Constructor, instantiates statsd reported to provided host and port. */
-    public StatsdReporter(String statsdHost, int statsdPort, boolean telemetry, int queueSize) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/datadog/jmxfetch/reporter/StatsdReporter.java
```diff
@@ -38,14 +38,15 @@ private void init() {
...
-                        + "queueSize={} entityId={}",
+        log.info("Initializing blocking Statsd reporter with parameters host={} port={} "
+                        + "telemetry={} queueSize={} entityId={}",
                 this.statsdHost, this.statsdPort, this.telemetry, this.queueSize, entityId);
         NonBlockingStatsDClientBuilder builder = new NonBlockingStatsDClientBuilder()
                 .hostname(this.statsdHost)
                 .port(this.statsdPort)
                 .enableTelemetry(this.telemetry)
                 .queueSize(this.queueSize)
+                .blocking(true)
```
remeh, line 49: It's a simply change and most likely the default we want, however, we have many Agents out there defaulting to the contrary for now. Because of that, should we consider having a configuration flag if for some reasons we would want to get back to non-blocking on some deployments?

vickenty (approved):

PR #423, opened 2023-04-06
Updates lombok dependency to remove need for tools.jar

scottopell:
I reviewed the [lombok changelog](https://github.com/projectlombok/lombok/blob/master/doc/changelog.markdown) and I don't see anything that should cause issues (tested this locally as well).
The lombok issue that discusses this fix is a bit of a mess, but [here](https://github.com/projectlombok/lombok/issues/2681) it is.

Newer JDKs don't ship with tools.jar from what I understand, this should make it easier to work on this project with a wider variety of java dev environments

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,13 +137,6 @@
             <version>${jcommander.version}</version>
         </dependency>
         <dependency>
-            <groupId>com.sun</groupId>
-            <artifactId>tools</artifactId>
-            <version>1.4.2</version>
-            <scope>system</scope>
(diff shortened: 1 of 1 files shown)


scottopell:
Can't remove this dependency, lombok is not the only codepath that utilizes `tools.jar` 
```
[ERROR] /home/circleci/project/src/main/java/org/datadog/jmxfetch/AttachApiConnection.java:[37,40] package com.sun.tools.attach does not exist
```

See https://github.com/DataDog/jmxfetch/blob/master/src/main/java/org/datadog/jmxfetch/AttachApiConnection.java#L38

PR #426, opened 2023-04-10
Bean subscription

scottopell:
Currently there is a periodic refresh of the beans (and their attributes) which can result in CPU spikes as these remote RMI calls are expensive.

This PR aims to implement a "subscription mode" for beans where a `javax.management.NotificationListener` is used to subscribe to bean registration/unregistration events.

This approach should spread out the expensive RMI calls over time and will reduce the total amount of calls needed. Each bean refresh empties out the current set of "known beans" and re-queries the MBeanServer for all matching beans and attributes. This creates an inefficiency when the monitored app does not frequently change the beans as each refresh cycle will repeat the same queries.

Can be enabled per-instance by adding `enable_bean_subscription: true` to the instance config.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -25,7 +25,7 @@ import javax.management.ObjectName;
 import javax.security.auth.login.FailedLoginException;
 
 @Slf4j
-public class Instance {
+public class Instance implements BeanListener{
     private static final List<String> SIMPLE_TYPES =
             Arrays.asList(
                     "long",
@@ -102,6 +102,7 @@ public class Instance {
     private AppConfig appConfig;
     private Boolean cassandraAliasing;
     private boolean emptyDefaultHostname;
+    private boolean enableBeanSubscription;
 
     /** Constructor, instantiates Instance based of a previous instance and appConfig. */
     public Instance(Instance instance, AppConfig appConfig) {
@@ -256,6 +257,8 @@ public class Instance {
         } else {
             log.info("collect_default_jvm_metrics is false - not collecting default JVM metrics");
         }
+        this.enableBeanSubscription = true;
+
     }
 
     public static boolean isDirectInstance(Map<String, Object> configInstance) {
@@ -416,6 +419,10 @@ public class Instance {
             throws IOException, FailedLoginException, SecurityException {
         log.info("Trying to connect to JMX Server at " + this.toString());
         connection = getConnection(instanceMap, forceNewConnection);
+        if (this.enableBeanSubscription) {
+            log.info("Subscribing for bean notifications before init");
+            this.subscribeToBeans();
+        }
         log.info(
                 "Trying to collect bean list for the first time for JMX Server at "
                         + this.toString());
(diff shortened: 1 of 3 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -506,176 +600,252 @@ public boolean timeToCollect() {
...
+        } finally {
+            log.debug("Bean unregistration processed. {}", mBeanName);
+            this.beanAndAttributeLock.unlock();
+        }
+    }
+
+
+    private void subscribeToBeans() {
+        List<String> beanScopes = this.getBeansScopes();
+        BeanSubscriber subscriber = new BeanSubscriber(beanScopes, this.connection, this);
```
scottopell, line 832: This may be more complicated than needed, in my testing I see that JMX MBean notifications come in from a `ClientNotif` thread that seems to be managed by jmx, my thought here was that the thread which subscribed to the notifications would be the one notified, but I don't think that assumption is correct.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -690,29 +860,61 @@ public List<String> getBeansScopes() {
...
-                        e);
             }
-        }
 
-        this.beans = (this.beans.isEmpty()) ? connection.queryNames(null) : this.beans;
-        this.lastRefreshTime = System.currentTimeMillis();
+            if (fullBeanQueryNeeded) {
+                newBeans = connection.queryNames(null);
+            }
+            if (this.beanSubscriptionActive && !fullBeanQueryNeeded && !isInitialQuery) {
```
scottopell, line 892: TODO, decide how to enable this extra audit logging or if some other form of auditing would be more useful.

scottopell:
- [ ] Can audit logging answer "Were there any beans missed?"
- [ ] Simplify bean subscription creation, can execute on main thread
- [ ] Add more tests around metric counting (ie, if we reach maxMetrics and new bean subscription arrives, do we properly ignore it? If we reach maxMetrics and a new bean unregistration arrives, do we properly decrement the metric count?)
- [ ] What is recommended configuration for internal testing?
- [ ] Add metrics around instance count, collection time, metrics collected, etc (basically jmxfetch telemetry)

gh123man (approved):
From a high level (I don't have deep knowledge of JMX) this LGTM. 
Just some nits and comments you can probably ignore 😅

src/main/java/org/datadog/jmxfetch/BeanNotificationListener.java
```diff
@@ -0,0 +1,59 @@
...
+    private final BeanTracker beanListener;
+
+    public BeanNotificationListener(final BeanTracker bl) {
+        this.beanListener = bl;
+        this.queue = new LinkedBlockingQueue<>();
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                while (true) {
+                    MBeanServerNotification mbs;
```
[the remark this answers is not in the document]
scottopell: No I don't, moved the decl inside the `try`

tools/misbehaving-jmx-server/README.md
```diff
@@ -27,6 +27,15 @@ a secondary `init` payload that contains the correct RMI Hostname. It is designe
 - `SUPERVISOR_PORT` - HTTP control port for the supervisor process (if using) (default 8088)
 - `MISBEHAVING_OPTS` - Manages memory, GC configurations, and system properties of the Java process running the JMXServer (default `-Xmx128M -Xms128M`)
 
+To enable debug logging, set the system property `-Dorg.slf4j.simpleLogger.defaultLogLevel=debug`.
+This can be done in the Dockerfile `CMD` if running in a container
+
+For extra debug logging around RMI systems, try any/all of the following:
```
apigirl, line 33:
```suggestion
For extra debug logging around RMI systems, try the following:
```

apigirl (commented):
editorial feedback

PR #429, opened 2023-04-12
Cleaning up App.java

carlosroman:
To make it a little bit easier to read I went through and tidied up `App.java`, mainly making some/all fields `final` so it would be more obvious where we are changing values. This has reduced the number of "warnings" looking at the file.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -72,44 +72,49 @@ public class App {
     private int lastJsonConfigTs;
     private Map<String, Object> adJsonConfigs;
     private Map<String, YamlParser> configs;
-    private Map<String, YamlParser> adPipeConfigs = new ConcurrentHashMap<String, YamlParser>();
-    private List<Instance> instances = new ArrayList<Instance>();
-    private Map<String, Instance> brokenInstanceMap = new ConcurrentHashMap<String, Instance>();
+    private Map<String, YamlParser> adPipeConfigs = new ConcurrentHashMap<>();
+    private List<Instance> instances = new ArrayList<>();
+    private Map<String, Instance> brokenInstanceMap = new ConcurrentHashMap<>();
     private AtomicBoolean reinit = new AtomicBoolean(false);
 
     private TaskProcessor collectionProcessor;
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -775,11 +779,16 @@ private void sendServiceCheck(
...
+            final Reporter reporter,
+            final String checkName,
+            final String serviceCheckPrefix,
+            final String status,
+            String message,
+            final String[] tags) {
+        final String serviceCheckName = String.format("%s.can_connect", serviceCheckPrefix);
 
-        if (status != Status.STATUS_ERROR) {
+        if (!status.equals(Status.STATUS_ERROR)) {
```
scottopell, line 791: !! good catch here

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -1001,46 +1010,47 @@ static TaskStatusHandler processCollectionResults(
...
                 // All was good, add instance
-                instances.add(instance);
-                log.info("Successfully initialized instance: " + instance.getName());
+                this.instances.add(instance);
+                log.info("Successfully initialized instance: {}", instance.getName());
             } catch (Throwable e) {
                 log.warn(
-                    "Could not initialize instance: " + instance.getName()
-                    + ": ", e);
+                    "Could not initialize instance: {}:", instance.getName(), e);
```
scottopell, line 1033:
```suggestion
                    "Could not initialize instance: {}: {}", instance.getName(), e);
```

src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -1167,12 +1178,12 @@ private <T> void processCollectionStatus(
 
             } finally {
 
-                if (instanceStatus == Status.STATUS_ERROR) {
+                if (Status.STATUS_ERROR.equals(instanceStatus)) {
```
scottopell, line 1181: 👍

scottopell (changes requested):
Found one nit, but looks good otherwise

scottopell (approved):

PR #431, opened 2023-04-14
Remove redundant API call

vickenty:
getMBeanInfo is the slowest part of the attribute matching process. Speed up the process by making the call once and reusing the info object to get everything we need.


scottopell (dismissed):

src/main/java/org/datadog/jmxfetch/Connection.java
```diff
@@ -18,6 +18,7 @@
 import javax.management.AttributeNotFoundException;
 import javax.management.InstanceNotFoundException;
 import javax.management.IntrospectionException;
+import javax.management.MBeanInfo;
 import javax.management.MBeanAttributeInfo;
 import javax.management.MBeanException;
```
carlosroman, line 23:
```suggestion
import javax.management.MBeanAttributeInfo;
import javax.management.MBeanInfo;
import javax.management.MBeanException;
```
Just to fix the CI linter test.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -20,6 +20,7 @@
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import javax.management.MBeanInfo;
 import javax.management.MBeanAttributeInfo;
```
carlosroman, line 24:
```suggestion
import javax.management.MBeanAttributeInfo;
import javax.management.MBeanInfo;
```
Again, just to fix CI.

carlosroman (approved):

PR #432, opened 2023-04-14
Avoid thread leak and data race in cleanUpAsync

vickenty:
A thread started in cleanUpAsync may overlap with Instance#init called from fixBrokenInstances, closing a new freshly created connection object instead of the old one. This is fixed by capturing specific connection object in the thread, not the whole instance.

If a connection issue persists over multiple collection intervals, there will be multiple calls to cleanUpAsync while an Instance references the same connection object. closeConnector may block to wait for a network protocol timeout; and, because the implementation is synchronized, multiple concurrent calls will wait for it sequentially, causing build-up of threads all waiting to close a connection. Reset the Instance connection to null, to make sure only one thread is trying to close a connection.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -776,29 +776,32 @@ public class Instance {
         this.appConfig = null;
         if (connection != null) {
             connection.closeConnector();
+            connection = null;
         }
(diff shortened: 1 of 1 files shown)


carlosroman (approved):

jk2l:
This changes may have caused #519

PR #436, opened 2023-04-28
Updates dummyTrustManager to use more supported security protocol

scottopell:
Some customers run JMXFetch on a host install where their own JVM is used. These JVMs may restrict the available security algorithms to exclude SSL which would lead to an error similar to this:
```
java.security.NoSuchAlgorithmException: SSL SSLContext not available
   at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:159)
   at java.base/javax.net.ssl.SSLContext.getInstance(SSLContext.java:168)
   at org.datadog.jmxfetch.HttpClient.<init>(HttpClient.java:91)
   at org.datadog.jmxfetch.App.<init>(App.java:115)
   at org.datadog.jmxfetch.JmxFetch.main(JmxFetch.java:56)
   at org.datadog.jmxfetch.App.main(App.java:91)
```

This PR changes our `dummyTrustManager` to use `TLS` instead of `SSL` (allowing the JVM to pick the specific version of TLS supported), which should have wider support across JVMs. 

Also adds an `error` log specifying the impact of a missing `dummyTrustManager` to help with future triage.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/HttpClient.java
+++ b/src/main/java/org/datadog/jmxfetch/HttpClient.java
@@ -26,7 +26,7 @@ public class HttpClient {
     // I found 'TLS' was the appropriate protocol (will use the latest support TLSv? version)
     // rather than specifically 'TLSv1' as the docs recommend
     // https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLContext.html
-    private static final String minRequiredSSLProtocol = "TLS";
+    private static final String sslProtocol = System.getProperty("jmxfetch.min_tls_version", "TLS");
 
     public static class HttpResponse {
         private int responseCode;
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/HttpClient.java
```diff
@@ -21,6 +22,11 @@ public class HttpClient {
     private int port;
 
     private static final String USER_AGENT = "Datadog/JMXFetch";
+    // Per javadocs, this is the only version that all compliant JVMs are required to support
+    // I found 'TLS' was the appropriate protocol (will use the latest support TLSv? version)
+    // rather than specifically 'TLSv1' as the docs recommend
+    // https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLContext.html
+    private static final String minRequiredSSLProtocol = "TLS";
```
vickenty, line 29: Would it make sense to make it configurable? Something like:

```suggestion
    private static final String minRequiredSSLProtocol = System.getProperty("jmxfetch.min_tls_version", "TLS");
```

PR #445, opened 2023-05-26
Adds error-prone bug checker as a compile step

scottopell:
Mostly just seeing if this compiles on older JDK versions (lots of notes about jdk 8 support in error-prone docs)


scottopell:
Not going to finish this, can revisit at some future point when we can drop support for older JVMs

PR #449, opened 2023-06-14
Avoiding app shutdown race condition

kyleboon:
When an application received SIGTERM, if the data dog agent has not completely finished it's initialization, adding the shutdown hook results in an exception. Catching the exception and stopping the initialization process should avoid this.

Example stack trace when this happens:

```
java.lang.IllegalStateException: Shutdown in progress
	at java.base/java.lang.ApplicationShutdownHooks.add(Unknown Source)
	at java.base/java.lang.Runtime.addShutdownHook(Unknown Source)
	at org.datadog.jmxfetch.App.run(App.java:154)
	at datadog.trace.agent.jmxfetch.JMXFetch$1.run(JMXFetch.java:131)
	at java.base/java.lang.Thread.run(Unknown Source)
```

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -155,8 +155,13 @@ public class App {
         // set up the config status
         this.appConfig.updateStatus();
 
-        // Adding another shutdown hook for App related tasks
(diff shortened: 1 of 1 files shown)


scottopell (approved):
Thanks for the contribution!

PR #450, opened 2023-06-14
TestContainer-based tests use container IP instead of port mapping

scottopell:
## What is changing:
1. When TestContainers unit tests start the `misbehaving-jmx-server`, they now communicate directly via the container's IP address instead of utilizing a mapped port on the docker-host.
2. The `misbehaving-jmx-server` jar now has 2 entrypoints, one is the existing misbehaving jmx server, and the second is a supervisor process which is responsible for starting the jmx server with the correct RMI hostname.
    - This introduces a 2-phase startup for the `misbehaving-jmx-server` as the JMX server _must_ have the correct RMI hostname set in order for the TestContainer client to connect successfully. That is the container's IP address, which can't be known until the container starts.
    - If you have a better idea for addressing the container to make this a one-stage startup, please point it out! This is the best I was able to think up.

## Why is this changing?
Short answer: for compatibility with gitlab CI environment.
Longer answer: In our gitlab environment, there are iptable rules preventing a container from communicating with the docker-host on anything except for the docker daemon port. The recommended TestContainers approach here is to communicate to spawned containers via the docker-host (ie, gateway of the bridge network) to support cases where the docker containers are spawned in a remote network that you don't have access to. In our case, we don't need to support that scenario as we actually require direct connectivity to support RMI, so this simplifies things a bit and adds support for more environments to run in.

## Diagram please?
Of course! 
image

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,9 +1,38 @@
 stages:
   - deploy_to_sonatype
   - create_key
+  - run_tests
 
 variables:
   REGISTRY: 486234852809.dkr.ecr.us-east-1.amazonaws.com
+  # ryuk is a privileged container started by testcontainers which cleans up containers at the end of testing
+  # It is not necessary if the docker env cleans itself up (which I'm assuming gitlab 'runner:docker' runners do)
+  TESTCONTAINERS_RYUK_DISABLED: "true"
+
+
+# Testing is handled by circleCI for PRs, but we also run maven tests as part of the deploy process
+# This run_tests job is useful for ensuring the tests run fine without any of the deployment bits
+run_tests:
+  stage: run_tests
+
+  rules:
+    - when: manual
+      allow_failure: true
+
+  tags:
+    - "runner:docker"
+
+  image: maven:3.9.1-eclipse-temurin-8
+
+  script:
+    - export TESTCONTAINERS_RYUK_DISABLED=true
+    - mvn -Dhttps.protocols=TLSv1.2 -Dcheckstyle.skip=true -Dtests.log_level=info -Djdk.attach.allowAttachSelf=true -B test
+
+  artifacts:
+    expire_in: 1 mos
+    paths:
+      - ./target/surefire-reports/*.txt
+
 
 # From the tagged repo, push the release artifact
 deploy_to_sonatype:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,7 +6,7 @@ stages:
 variables:
   REGISTRY: 486234852809.dkr.ecr.us-east-1.amazonaws.com
   # ryuk is a privileged container started by testcontainers which cleans up containers at the end of testing
-  # It is not necessary if the docker env cleans itself up (which I'm assuming gitlab 'runner:docker' runners do)
+  # It is not necessary for our gitlab CI env as #ci-cd infra tears everything down at the end of a gitlab run
   TESTCONTAINERS_RYUK_DISABLED: "true"
 
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,7 +1,7 @@
 stages:
   - deploy_to_sonatype
   - create_key
-  - run_tests
+  - run_unit_tests
 
 variables:
   REGISTRY: 486234852809.dkr.ecr.us-east-1.amazonaws.com
@@ -11,9 +11,9 @@ variables:
 
 
 # Testing is handled by circleCI for PRs, but we also run maven tests as part of the deploy process
-# This run_tests job is useful for ensuring the tests run fine without any of the deployment bits
-run_tests:
-  stage: run_tests
+# This run_unit_tests job is useful for ensuring the tests run fine without any of the deployment bits
+run_unit_tests:
+  stage: run_unit_tests
 
   rules:
     - when: manual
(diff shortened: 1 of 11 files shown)


.gitlab-ci.yml
```diff
@@ -1,9 +1,38 @@
...
+    - when: manual
+      allow_failure: true
+
+  tags:
+    - "runner:docker"
+
+  image: maven:3.9.1-eclipse-temurin-8
+
+  script:
+    - export TESTCONTAINERS_RYUK_DISABLED=true
```
scottopell, line 28:
```suggestion
```

.gitlab-ci.yml
```diff
@@ -1,9 +1,38 @@
...
+    - mvn -Dhttps.protocols=TLSv1.2 -Dcheckstyle.skip=true -Dtests.log_level=info -Djdk.attach.allowAttachSelf=true -B test
```
scottopell, line 29: Add the necessary options (I think just `allowAttachSelf`) to the sonatype deploy job

src/test/java/org/datadog/jmxfetch/JMXServerSupervisorClient.java
```diff
@@ -0,0 +1,81 @@
...
+    public JMXServerSupervisorClient(String host, int port) {
+        this.host = host;
+        this.port = port;
+    }
+
+    public void setRmiHostname(String rmiHostname) throws IOException {
+        sendPostRequestWithFormParams("/config", Collections.singletonMap("rmiHostname", rmiHostname));
+    }
+
+    private void sendPostRequestWithFormParams(String endpoint, Map<String, String> formParams) throws IOException {
```
scottopell, line 29: consolidate this into json body, don't duplicate the post request sender with one version for json body and a second for form params.

src/test/java/org/datadog/jmxfetch/TestReconnectContainer.java
```diff
@@ -47,24 +57,47 @@ private static boolean isDomainPresent(String domain, MBeanServerConnection mbs)
...
+    @Rule(order = 0)
     public GenericContainer<?> cont = new GenericContainer<>(img)
-        .withExposedPorts(rmiPort, controlPort)
+        .withExposedPorts(rmiPort, controlPort, supervisorPort)
         .withEnv(Collections.singletonMap("RMI_PORT", "" + rmiPort))
         .withEnv(Collections.singletonMap("CONTROL_PORT", "" + controlPort))
-        .waitingFor(Wait.forLogMessage(".*IAMREADY.*", 1));
+        .withEnv(Collections.singletonMap("SUPERVISOR_PORT", "" + supervisorPort))
+        .waitingFor(Wait.forLogMessage(".*JMX server has started.*", 1));
+        // TODO switch to a http ready check, but need to be able to address container by container IP, not gateway/docker-host
```
scottopell, line 68: address this, if its possible then do it, otherwise double check that the log will not be triggered before the JMX server is actually ready.

tools/misbehaving-jmx-server/src/main/java/org/datadog/supervisor/App.java
```diff
@@ -0,0 +1,165 @@
...
+                ctx.status(200);
+            } else {
+                ctx.status(500);
+            }
+        });
+
+        log.info("Supervisor HTTP control interface at :{}", App.config.controlHttpPort);
+        app.start(App.config.controlHttpPort);
+
+        log.info("Starting JMX subprocess");
```
scottopell, line 84: Consider _not_ starting the jmx subprocess until we recieve our initial bundle of init values (ie, rmiHostname).

This is currently attempting to support a case where some users of `misbehaving-jmx-server` may need the custom rmi hostname and some may not, but in reality all of our users _do_ need this support, so lets bake that assumption in and not try to be too smart

.gitlab-ci.yml
```diff
@@ -44,7 +72,7 @@ deploy_to_sonatype:
     - set -x
 
     - echo "Building release..."
-    - mvn -DperformRelease=true -Daether.checksums.algorithms=SHA-512,SHA-256,SHA-1,MD5 --settings ./settings.xml clean deploy
+    - mvn -Djdk.attach.allowAttachSelf=true -DperformRelease=true -Daether.checksums.algorithms=SHA-512,SHA-256,SHA-1,MD5 --settings ./settings.xml clean deploy
```
scottopell, line 75: Will be necessary in the future if we upgrade the JVM that we perform the releases under to 9+, but currently not needed. Just future-proofing here.

src/test/java/org/datadog/jmxfetch/JMXServerSupervisorClient.java
```diff
@@ -0,0 +1,57 @@
...
+        this.port = port;
+    }
+
+    public void initializeJMXServer(String rmiHostname) throws IOException {
+        String endpoint = "/init";
+        String jsonPayload = "{\"rmiHostname\": \""  + rmiHostname + "\"}";
+        sendPostRequestWithPayload(endpoint, jsonPayload);
+    }
+
+    private void sendPostRequestWithPayload(String endpoint, String jsonPayload) throws IOException {
```
scottopell, line 28: This is duplicated from `JMXServerControlClient.java`. I'll extract this to a base-class

scottopell:
ReTested MacOS and it still does not work with a naive `mvn test`, but still _does_ work using the docker-wormhole settings described in the README

carlosroman (approved):
We should add a card somewhere to track fixing testcontainer tests on macOS.

scottopell:
I don't think there's any way to fix the testcontainer tests on macos, its a limitation of the networking in docker-desktop.
The workaround in the readme works both with and without this PR

PR #453, opened 2023-07-06
custom jmxfetch bean to track number of ingested beans, attributes, and metrics

cmetz100:
This pr is part of a multi stage process to increase observability within jmxfetch. Another [pr](https://github.com/DataDog/jmxfetch/pull/467) adds a check that uses proccess_regex to attach to jmxfetch's jvm in order to collect mbean data from our own application. In this pr I have added jmx MBeans that are registered for each instance of jmxfetch and track the number of beans, metrics, and attributes that each instance collects and exposes this under the jmxFetch domain where the check from my other pr will be able to pick up this data in addition to general jmfetch metrics.


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,6 +266,19 @@ public Instance(
         } else {
             log.info("collect_default_jvm_metrics is false - not collecting default JVM metrics");
         }
+        //create bean to track data of this instance
+        JMXbean = beanManager.createJMXBean(this);
```
scottopell, line 270: This uses a static instance of the `BeanManager` class to create each telemetry bean, but I'm not sure why we need the static instance and the "factory" like aspect of this.

Is there a reason why we cannot simply create a new instance of the "telemetry bean" from `Instance` directly and set it as an instance property?

This would remove the need for `BeanManager` entirely and simplify the code structure here

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,6 +266,19 @@ public Instance(
...
+    }
+
+    public void updateBeanInfo()throws Exception {
+        if (JMXbean != null){
```
scottopell, line 274: Is it possible for `JMXbean` to be `null`?

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,6 +266,19 @@ public Instance(
...
+            refreshBeansList();
```
scottopell, line 275: This and the below `getMatchingAttributes` are expensive operations, if the instance is monitoring a JVM exposing 10s of thousands of beans, then we must go fetch each of those 10s of thousands of beans and check each attribute to see if it must be monitored.

In the current draft of this PR, we're calling `updateBeanInfo` in the main collect loop every 15s, so this will create a LOT more JMX traffic and use up more CPU.

We should try to structure the updates to the `beanCount` `attributeCount` and `metricCount` to use the existing bean refresh, existing attribute match, and existing metric collection instead of calling them again here.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,6 +266,19 @@ public Instance(
...
+            getMatchingAttributes();
+            JMXbean.setBeanCount(beans.size());
+            JMXbean.setAttributeCount(matchingAttributes.size());
+            JMXbean.setMetricCount(getMetrics().size());
```
scottopell, line 279: Similar to my comment above, `getMetrics` is a very expensive call, we should not be adding more calls to it.

src/main/java/org/datadog/jmxfetch/util/BeanManager.java
```diff
@@ -0,0 +1,53 @@
...
+        registeredBeans = new HashMap<>();
+    }
+
+    private ObjectName getObjName(String domain,String name) throws MalformedObjectNameException {
+        return new ObjectName(domain + ":name=" + name);
+    }
+
+    public BeanJavaApp createJMXBean(Instance instance){
+        //if the instance already has a bean just return that 
+        if (registeredBeans.get(instance.getCheckName()) != null){
```
scottopell, line 34: What is a scenario when an instance already has a telemetry bean?

scottopell (changes requested):
This is a good start in that it works, but now we need to make it efficient. We should not be creating any un-necessary work to update our telemetry, instead we should try to use the existing data that is already collected as part of JMXFetch's execution

scottopell:
Also please try to add some unit tests for this. You should be able to get something working using the `TestApp` from the existing unit tests and assert on the values from the instance's telemetry bean.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,8 +271,50 @@ public Instance(
...
+
+    public String escape(String inputString){
+        final String[] metaCharacters = {"*","?"};
+    
+        for (int i = 0 ; i < metaCharacters.length ; i++){
+            if(inputString.contains(metaCharacters[i])){
+                inputString = inputString.replace(metaCharacters[i],"\\"+ metaCharacters[i]);
+            }
+        }
+        inputString = "\"" + inputString + "\"";
```
scottopell, line 293: lets try to use the `ObjectName.quote` where its possible, rather than manually quoting/escaping/unquoting

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,8 +271,50 @@ public Instance(
...
+        JmxfetchTelemetry bean = new JmxfetchTelemetry(this);
+        log.info("Created jmx bean for instance: " + this.getCheckName());
+
+     try {
+                JMXbean_name = getObjName("JMXFetch" , this.getName());
+                mbs.registerMBean(bean,JMXbean_name);
+                log.info("Succesfully registered jmx bean for instance: " + this.getCheckName());
+
+        } catch (MalformedObjectNameException | InstanceAlreadyExistsException | MBeanRegistrationException | NotCompliantMBeanException e) {
+                    log.warn("Could not register bean for instance: " + this.getCheckName());
```
scottopell, line 309: make sure you print which type of exception, I believe you're only printing out the stack trace at the moment

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -485,6 +541,16 @@ public List<Metric> getMetrics() throws IOException {
...
             }
         }
+        JMXbean.setBeanCount(beans.size());
+        JMXbean.setAttributeCount(matchingAttributes.size());
+        JMXbean.setMetricCount(metrics.size());
+        log.info("There are " + mbs.getMBeanCount() + "total beans currrently registered");
+        try {
+            log.info("updating jmx bean for instance" + this.getName() +  " with bean: " + mbs.getObjectInstance(JMXbean_name).toString());
+        } catch (Exception e) {
+            // TODO: handle exception
```
scottopell, line 551: ensure we fill this in eventually

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -706,7 +772,7 @@ private void refreshBeansList() throws IOException {
         }
 
         this.beans = (this.beans.isEmpty()) ? connection.queryNames(null) : this.beans;
-        this.lastRefreshTime = System.currentTimeMillis();
+        this.lastRefreshTime = System.currentTimeMillis();
```
scottopell, line 775: whitespace at the end of the line, look up how to configure VSCode to automatically trim trailing whitespace

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -773,8 +839,18 @@ public boolean isLimitReached() {
         return this.limitReached;
     }
 
+    private void cleanBean(){
```
scottopell, line 842:
```suggestion
    private void cleanupTelemetryBean(){
```

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,8 +271,35 @@ public Instance(
...
+        log.debug("Created jmx bean for instance: " + this.getCheckName());
+
+        try {
+            jmxBeanName = getObjName("JMXFetch" , this.getName());
+            mbs.registerMBean(bean,jmxBeanName);
+            log.debug("Succesfully registered jmx bean for instance: " + this.getCheckName());
+
+        } catch (MalformedObjectNameException | InstanceAlreadyExistsException
+                | MBeanRegistrationException | NotCompliantMBeanException e) {
+            log.debug("Could not register bean for instance: " + this.getCheckName(),e);
```
scottopell, line 295:
```suggestion
            log.warn("Could not register bean for instance: " + this.getCheckName(),e);
```

lets up the visibility of this error a bit, as far as we know this should never occur, so if it does, its worth investigating more.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -257,8 +271,35 @@ public Instance(
...
+            e.printStackTrace();
```
scottopell, line 296:
```suggestion
```

no longer needed since you're now printing the entire `e` exception above

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -485,6 +526,13 @@ public List<Metric> getMetrics() throws IOException {
                 }
             }
         }
+        jmxBean.setBeanCount(beans.size());
+        jmxBean.setAttributeCount(matchingAttributes.size());
+        jmxBean.setMetricCount(metrics.size());
+        log.debug("Updated jmx bean for instance: " + this.getCheckName() + " for bean: "
+                + jmxBean.toString() + " With number of beans = " + jmxBean.getBeanCount()
```
scottopell, line 533: what does `jmxBean.toString()` look like here? Do you have an example of this log line?

scottopell (commented):
Mostly just naming comments

scottopell (approved):

PR #456, opened 2023-07-18
Caleb/jmxfetch mbean remove quotes config

cmetz100:


src/main/java/org/datadog/jmxfetch/JmxAttribute.java
```diff
@@ -191,6 +194,22 @@ private List<String> getBeanParametersList(
         return beanTags;
     }
 
+    private List<String> normalizeInstanceParameters(List<String> beanParametersList) {
```
scottopell, line 197: Can we consolidate this functionality into the existing `sanitizeParameters` function? They're both doing similar things at a high level -- iterating through a list of bean parameters and formatting them into tags.

src/main/java/org/datadog/jmxfetch/JmxAttribute.java
```diff
@@ -191,6 +194,22 @@ private List<String> getBeanParametersList(
...
+        List<String> instanceTagsList = new ArrayList<String>(beanParametersList.size());
+        for (String beanParameter : beanParametersList) {
+            if (normalizeBeanParamTags == true) {
+                String oldBeanParamater = beanParameter;
```
scottopell, line 201:
```suggestion
                String oldBeanParameter = beanParameter;
```

src/main/java/org/datadog/jmxfetch/JmxAttribute.java
```diff
@@ -191,6 +194,22 @@ private List<String> getBeanParametersList(
...
+                beanParameter = beanParameter.replace("\"","");
+                if (!oldBeanParamater.equals(beanParameter)) {
+                    log.info("Removing quotes on bean with parameter: " + oldBeanParamater);
```
scottopell, line 204: info is too high a level for this. Maybe `debug` level, but even that may be too high. This will happen a LOT and the logs are not that useful. I'd be inclined to get rid of this log altogether.

scottopell (commented):
Good change overall, big picture I'd like to avoid adding a new function similar to the existing `sanitizeParameterNames`, but other than that its a good change and good tests.

src/main/java/org/datadog/jmxfetch/JmxAttribute.java
scottopell: we discussed this, `sanitizeParameters` is `static` which is why this was separated. Lets try to make `sanitizeParameters` non-static and consolidate.

src/main/java/org/datadog/jmxfetch/JmxAttribute.java
```diff
@@ -195,12 +198,17 @@ private List<String> getBeanParametersList(
...
-    private static List<String> sanitizeParameters(List<String> beanParametersList) {
+    private List<String> sanitizeParameters(List<String> beanParametersList) {
         List<String> defaultTagsList = new ArrayList<String>(beanParametersList.size());
         for (String rawBeanParameter : beanParametersList) {
             // Remove `|` characters
             String beanParameter = rawBeanParameter.replace("|", "");
 
+            // Remove `"` characters
+            if(normalizeBeanParamTags == true) {
+                beanParameter = beanParameter.replace("\"","");
```
scottopell, line 209: do you think we should remove all quotes or only the quotes at the start/end of the bean parameter?

src/main/java/org/datadog/jmxfetch/JmxAttribute.java
```diff
@@ -191,16 +194,31 @@ private List<String> getBeanParametersList(
         return beanTags;
     }
 
+    private String unquote(String beanParameter) {
```
scottopell, line 197: Lets add a brief comment describing the try-catch here. Something to help future readers understand that if the `beanParameter` doesn't contain quotes, its expected to hit the catch and return the original.

scottopell (approved):
Looks good!

scottopell:
Can you just update the title of this PR and the description to describe why this is useful? When we squash+merge, the PR title and description are in the commit msg

PR #457, opened 2023-07-21
Update management agent logic and comments for java 7 vs 8 vs 9

cmetz100:
This pr updates some minor logic and comments regarding the state of the localManagementAgent across java versions and its role in attaching to a local running java process. The goal is to have a better understanding of how we can use process_regex to configure a jmxfetch check that attaches to itself.


scottopell:
This looks pretty good to me, can you take it out of draft and add a description to this PR?

scottopell:
Oh see also the linter warnings. We enforce a 100char line length in this repo, you have some violations.

PR #459, opened 2023-07-24
updated tests and misbehaving to work on MacOS

cmetz100:


tools/misbehaving-jmx-server/src/main/java/org/datadog/misbehavingjmxserver/App.java
```diff
@@ -83,6 +83,9 @@ public static void main( String[] args ) throws IOException, MalformedObjectName
         // I don't think this call is actually important for jmx, the below 'env' param to JMXConnectorServerFactory is the important one
         RMISocketFactory.setSocketFactory(customRMISocketFactory);
 
+        // Explicitly set rmi server host, seems to be issues with the default on our corp computers host MacOS
```
scottopell, line 86:
```suggestion
        // Explicitly set RMI hostname to specified argument value
```

I don't think its necessary to mention the specific case we're working around.  Really this way is the most correct and it should have been written this way in the first place. So you're just correcting my mistake :)

pom.xml
```diff
@@ -223,7 +223,7 @@
                 <version>${maven-surefire-plugin.version}</version>
                 <configuration>
                     <!-- required for tests in jdk9+ ref https://bugs.openjdk.org/browse/JDK-8180425 -->
-                    <argLine>-Djdk.attach.allowAttachSelf=true</argLine>
+                    <argLine>-Djdk.attach.allowAttachSelf=true -Djava.rmi.server.hostname=localhost</argLine>
```
scottopell, line 226: add a comment above explaining that this is needed in some macos configurations in order to be able to connect.

scottopell (approved):

scottopell (approved):

PR #463, opened 2023-08-03
Adding Maven Wrapper to project

carlosroman:
Updated all scripts/pipelines to use the Maven Wrapper. This should make it easier to get started when working on this project as all you need is a JDK. 

Currently package the app fails on modern JDKs so you need to build/package it using:
```
docker run -it --rm --name my-maven-project -v "$(pwd)":/usr/src/app -w /usr/src/app eclipse-temurin:8u382-b05-jdk ./mvnw -DskipTests clean package
```


scottopell (approved):

brett0000FF (approved):
Looks good! Just a few small suggestions.

scottopell (approved):

PR #481, opened 2023-09-28
Adding some composite data tests

scottopell:
This make a few changes:

1. Moves the "is attribute of type XYZ" check inside each attribute class
2. Adds a test case with both a simple compositedata attribute and a nested compositedata attribute (the latter is not supported)
3. Small exception handler cleanup with telemetry bean registration.

I wanted to change the "is attribute of type XYZ" check to use a more dynamic method to avoid having to list all possible subclasses (ref `TabularData` vs `TabularDataSupport`).

Unfortunately this is not an easy change due to the fact that `TabularDataSupport` actually implements `Map`, so TabularDataSupport was getting picked up as a `JmxComplexAttribute` vs a `JmxTabularAttribute`.
Change was attempted in https://github.com/DataDog/jmxfetch/commit/baa60df420751fd384f7113280514b909173f8c3

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -38,44 +38,6 @@ import javax.security.auth.login.FailedLoginException;
 
 @Slf4j
 public class Instance {
-    private static final List<String> SIMPLE_TYPES =
-            Arrays.asList(
-                    "long",
-                    "java.lang.String",
-                    "int",
-                    "float",
-                    "double",
-                    "java.lang.Double",
-                    "java.lang.Float",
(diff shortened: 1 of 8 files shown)


src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -295,13 +257,17 @@ private InstanceTelemetry createJmxBean() {
         try {
             instanceTelemetryBeanName = getObjName(appConfig.getJmxfetchTelemetryDomain(),
                      this.getName());
+        } catch (MalformedObjectNameException e) {
```
scottopell, line 260: These are unrelated changes, I'll remove them from this PR

PR #484, opened 2023-10-11
Add warning for duplicate instance name

rayz:
We currently attempt to instantiate all instances even if an instance with the same name has already been seen.

This pr adds a warn log

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/App.java
+++ b/src/main/java/org/datadog/jmxfetch/App.java
@@ -29,11 +29,13 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


src/main/java/org/datadog/jmxfetch/App.java
```diff
@@ -893,6 +901,11 @@ public void init(final boolean forceNewConnection) {
                 final String checkName = (String) checkConfig.get("check_name");
                 for (Map<String, Object> configInstance : configInstances) {
                     log.info("Instantiating instance for: " + checkName);
+                    final String instanceName = (String) configInstance.get("name");
```
scottopell, line 904: this could return `null`, what would happen to your code if it does?

rayz: made new commit to add a `null` check

scottopell (approved):

PR #485, opened 2023-10-11
Clean up instance telemetry bean in tests

rayz:
this pr:
1. adds a cleanup for instances and their telemetry bean
2. renames some of the instance names in our test yaml so that they are unique

now, when you run `mvn test -Dtest=TestApp | grep InstanceAlreadyExistsException` , there are no exceptions

### DIFF
--- a/src/test/java/org/datadog/jmxfetch/TestCommon.java
+++ b/src/test/java/org/datadog/jmxfetch/TestCommon.java
@@ -110,6 +110,16 @@ public class TestCommon {
         }
     }
 
+    /**
+     * Clear instances and their instance telemetry bean after execution of every test.
+     */
+    @After
+    public void clearInstances() {
+        if (app != null) {
+            app.clearAllInstances();
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


scottopell (approved):

src/test/java/org/datadog/jmxfetch/TestCommon.java
```diff
@@ -110,6 +110,16 @@ public void unregisterMBean() throws MBeanRegistrationException, InstanceNotFoun
         }
     }
 
+    /**
+     * Clear instances and their instance telemetry bean after execution of every test.
+     */
+    @After
+    public void clearInstances() {
+        if (app != null) {
+            app.clearInstances(getInstances());
```
scottopell, line 119: Lets make `protected void clearAllInstances()` instead of exposing `clearInstances` as `protected`. This can wrap up the call to `getInstances` as well.

Motivation here is to keep `clearInstances` as private and expose a more stable `protected` interface.

scottopell (approved):

PR #487, opened 2023-10-17
Add JMXFetch telemetry around bean/attribute matching

rayz:
these changes:

1. add telemetry for the ratio of beans that had at least one attribute match
2. add telemetry for  # of wildcard bean queries

additional notes:
a following pr in the datadog-agent repo is needed to display these stats onto the `datadog-agent status -v` output


src/main/java/org/datadog/jmxfetch/util/InstanceTelemetry.java
```diff
@@ -7,12 +7,18 @@ public class InstanceTelemetry implements InstanceTelemetryMBean {
...
+    private int wildcardQueryCount;
+    private double attributeMatchRatio;
 
     /** Jmxfetch telemetry bean constructor. */
     public InstanceTelemetry() {
         beansFetched = 0;
         topLevelAttributeCount = 0;
         metricCount = 0;
+        domainsQueried = 0;
+        wildcardQueryCount = 0;
```
scottopell, line 20:
```suggestion
        wildcardDomainQueryCount = 0;
```

src/main/java/org/datadog/jmxfetch/util/InstanceTelemetry.java
```diff
@@ -7,12 +7,18 @@ public class InstanceTelemetry implements InstanceTelemetryMBean {
...
+        attributeMatchRatio = 0.0;
```
scottopell, line 21:
```suggestion
        beanMatchRatio = 0.0;
```

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -733,14 +752,24 @@ private void refreshBeansList() throws IOException {
                     ObjectName name = new ObjectName(scope);
                     this.beans.addAll(connection.queryNames(name));
                 }
-            } catch (Exception e) {
+                if (instanceTelemetryBean != null) {
+                    instanceTelemetryBean.setDomainsQueried(beanScopes.size());
+                }
+            } catch (MalformedObjectNameException e) {
```
scottopell, line 758: `setDomainsQueried` can throw a `MalformedObjectNameException`? 

Also since you catch both the exception types, I don't think think they can escape this method, so you should be able to remove the `throws` decl from the method signature I believ.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -733,14 +752,24 @@ private void refreshBeansList() throws IOException {
                     ObjectName name = new ObjectName(scope);
                     this.beans.addAll(connection.queryNames(name));
                 }
-            } catch (Exception e) {
+                if (instanceTelemetryBean != null) {
+                    instanceTelemetryBean.setDomainsQueried(beanScopes.size());
```
scottopell, line 756: Does one scope correspond to one domain?
We may want to unlink these from each other, presumably `getBeansScopes` could return scopes that are more specific than a domain, even if today all it does is return the domain.

Or rename `getBeansScopes` to `getBeansDomains` if that is indeed all it does.

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -434,7 +434,8 @@ public Connection getConnection(
 
     /** Initializes the instance. May force a new connection.. */
     public void init(boolean forceNewConnection)
-            throws IOException, FailedLoginException, SecurityException {
+            throws IOException, FailedLoginException, SecurityException, 
+                    MalformedObjectNameException {
```
scottopell, line 438: is this necessary?

scottopell (changes requested):
Some naming suggestions and clarifying questions, but this looks good overall

src/main/java/org/datadog/jmxfetch/Instance.java
rayz: removed

src/main/java/org/datadog/jmxfetch/Instance.java
rayz: `new ObjectName` can throw `MalformedObjectNameException`. removed the `throws` decl

rayz: L752
https://github.com/DataDog/jmxfetch/blob/b9671c319a24bad27e4e6b07b30fd7d8c73ba886/src/main/java/org/datadog/jmxfetch/Instance.java#L749-L752

scottopell (changes requested):
After discussion, the number of domains queried is not straightforward to instrument. Lets not collect it as telemetry.

scottopell (approved):

PR #488, opened 2023-10-19
Improves telemetry bean exception handling

scottopell:
Each `Instance` creates and manages a `Telemetry` bean. This PR improves the error messages when the telemetry bean cannot be registered for whatever reason.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -279,7 +279,7 @@ public class Instance {
             log.info("collect_default_jvm_metrics is false - not collecting default JVM metrics");
         }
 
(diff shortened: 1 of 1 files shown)


rayz (approved):

rayz:
https://github.com/DataDog/jmxfetch/blob/69cfd181466a2649e1593ff79971c5210dfbd5c4/src/main/java/org/datadog/jmxfetch/Instance.java#L550-L557

would it make sense to have a `null` check here?

scottopell:
re https://github.com/DataDog/jmxfetch/pull/488#issuecomment-1771507765

In the current code structure it is impossible for the telemetry bean to be null. Even if there is an error while registering, the bean itself is still returned and set on the instance.
If you'd like to submit a PR with a null check I'd approve it, but its not necessary IMO

PR #490, opened 2023-10-24
Attempt to fix misbehaving-jmx-server build

carlosroman:
Fixing the build of `misbehaving-jmx-server`. The issues comes from the version on Lombok and the JDK. Have pinned the JDK to be `17` as that one works with the version of Lombok being used. The error I was seeing pointed to this being the problem:

```
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project misbehavingjmxserver: Fatal error compiling: java.lang.NoSuchFieldError: Class com.sun.tools.javac.tree.JCTree$JCImpo
rt does not have member field 'com.sun.tools.javac.tree.JCTree qualid' -> [Help 1]
```

Also:

- Switched to using [Eclipse Temurin](https://projects.eclipse.org/projects/adoptium.temurin) as former AdoptOpenJDK project has moved to Eclipse Adoptium.
- Added [Maven Wrapper](https://maven.apache.org/wrapper/) to `misbehaving-jmx-server` as it made it easier to pin the JDK.


scottopell (approved):
🙇‍♂️

PR #491, opened 2023-10-25
AMLII-1194 - Updated CI JDKs to use the current LTS versions of the JDK

carlosroman:


.circleci/config.yml
```diff
@@ -17,34 +17,51 @@ jobs:
     docker:
       - image: cimg/openjdk:8.0
     <<: *lint_steps
+
+  # JDK 8 (LTS) - 31 Dec 2030 (extended support)
```
scottopell, line 21: If there's a good reference you have for these LTS dates, may be useful to drop in here

.circleci/config.yml
```diff
@@ -17,34 +17,51 @@ jobs:
...
-      - image: cimg/openjdk:15.0
-    <<: *build_steps
+
+  # JDK 11 (LTS) - 30 Sep 2029 (extended support)
   test_openjdk17:
     docker:
       - image: cimg/openjdk:17.0
     <<: *build_steps
+
+  # Available in Ubuntu 22.04
```
scottopell, line 39: What does this comment mean? Is jdk 18 what ships by default with ubuntu 22.04?

.circleci/config.yml
carlosroman: I used https://endoflife.date/oracle-jdk to work them out.

.circleci/config.yml
```diff
@@ -17,34 +17,51 @@ jobs:
...
   test_openjdk11:
     docker:
       - image: cimg/openjdk:11.0
     <<: *build_steps
-  test_openjdk15:
-    docker:
-      - image: cimg/openjdk:15.0
-    <<: *build_steps
+
+  # JDK 11 (LTS) - 30 Sep 2029 (extended support)
```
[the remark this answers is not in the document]
carlosroman: Yep 👍🏽

.circleci/config.yml
carlosroman: I was looking at which ones we test with + which ones are in the LTS version of Ubuntu. Mainly as trying to figure out what the test matrix should look like.

PR #492, opened 2023-10-31
AMLII-1173 - Modify JMXFetch to pass back a tag with the jmx_check_name to the Agent

rayz:
This pr:
1. Adds `jmx_check_name` tag to be passed to the agent which will be used by the Agent to determine metric origins
2. Updates tests to account for the additional tag (\# of tags += 1)
3. Rename some files wilth `.yml` to be `.yaml`. I found that `checkName` included `.yml` in it while files with `.yaml` did not. ex: the `jmx_additional_tags` test was displaying `jmx_check_name:jmx_additional_tags.yml` beforehand


src/test/java/org/datadog/jmxfetch/TestCommon.java
```diff
@@ -270,7 +270,7 @@ public void assertMetric(
                     assertEquals(countTags, mTags.size());
                 }
                 for (String t : tags) {
-                    assertTrue(mTags.contains(t));
+                    assertTrue(t + " missing from metric tags", mTags.contains(t));
```
carlosroman, line 273:
```suggestion
                     assertThat(mTags, hasItem(t));
```

You'll have to add the following to you imports:

```
import static org.hamcrest.CoreMatchers.hasItem;
import static org.hamcrest.MatcherAssert.assertThat;
```

scottopell (approved):

carlosroman (approved):

PR #497, opened 2023-11-10
Make misbehaving-jmx-server more deterministic

rayz:
This pr adds a new flag `--rng-seed` that is used to set the seed for the RNG used in `RandomIdentifier`
Also adds  flag `--seed-config-path` to specify config file to read `seed` from with the default being     `./misbehaving-seed-config.yaml`.

Seed configuration file overrides `--rng-seed` flag

Currently tested by

1. Run the misbehaving server `java -jar target/misbehavingjmxserver-1.0-SNAPSHOT-jar-with-dependencies.jar `
2. Run `curl localhost:8080/beans/test_1` `curl localhost:8080/beans/test_2`. 
3. Restart server
4. Run curl commands again and see that they are the same beans as the first run
5. Exit server and run with `--rng-seed` with any value
6. Run curl commands and see that the beans are different to original values with default seed


scottopell (commented):
Is this option supported when reading from a config file as well?

Change looks good overall

rayz:
> Is this option supported when reading from a config file as well?
Nope, will look into that now!

tools/misbehaving-jmx-server/src/main/java/org/datadog/misbehavingjmxserver/App.java
```diff
@@ -45,13 +45,20 @@ class AppConfig {
...
+    public Long rngSeed = 54321L;
+
     // Can only be set via env var
     public int controlPort = Defaults.JMXSERVER_CONTROL_PORT;
 
     @Parameter(names = {"--config-path", "-cfp"})
     public String config_path = "./misbehaving-jmx-domains-config.yaml";
 
+    @Parameter(names = {"--seed-config-path", "-scfp"})
+    public String seed_config_path = "./misbehaving-seed-config.yaml";
```
scottopell, line 58: Can we integrate this into the existing config file? I don't see a need to separate it into its own config file

scottopell (changes requested):

tools/misbehaving-jmx-server/src/main/java/org/datadog/misbehavingjmxserver/App.java
rayz: The old config file was specific to `domains`. I have made changes to only use 1 config file, renamed some of the codenames and just call config file `misbehaving-config.yaml` now. 

Open to suggestions for naming

scottopell (approved):

PR #502, opened 2023-12-27
AMLII-1358 - Adding support for Generational ZGC

carlosroman:
Starting with JDK 21, multi-generation support has been added to ZGC (see [here](https://inside.java/2023/10/09/sip084/)).

This PR adds a new mappings to the new GC default JMX metrics and creates the following metrics:

- two `jvm.gc.zgc_cycles_collection_count` - each with different `name` tags (`ZGC Major Cycles` and `ZGC Minor Cycles`)
- two `jvm.gc.zgc_cycles_collection_time` - each with different `name` tags (`ZGC Major Cycles` and `ZGC Minor Cycles`)
- two `jvm.gc.zgc_pauses_collection_count` - each with different `name` tags (`ZGC Major Pauses` and `ZGC Pauses Cycles`)
- two `jvm.gc.zgc_pauses_collection_time` - each with different `name` tags (`ZGC Pauses Cycles` and `ZGC Pauses Cycles`)

Example:
image

### DIFF
--- a/src/main/resources/org/datadog/jmxfetch/new-gc-default-jmx-metrics.yaml
+++ b/src/main/resources/org/datadog/jmxfetch/new-gc-default-jmx-metrics.yaml
@@ -147,3 +147,49 @@
       CollectionTime:
         alias: jvm.gc.zgc_pauses_collection_time
         metric_type: counter
+
+# Z Garbage Collector with ZGenerational
+- include:
+    domain: java.lang
+    type: GarbageCollector
+    name: ZGC Major Cycles
+    attribute:
+      CollectionCount:
+        alias: jvm.gc.zgc_cycles_collection_count
+        metric_type: counter
+      CollectionTime:
+        alias: jvm.gc.zgc_cycles_collection_time
+        metric_type: counter
+- include:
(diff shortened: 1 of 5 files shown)


apigirl (commented):
No docs team review needed

src/main/resources/org/datadog/jmxfetch/new-gc-default-jmx-metrics.yaml
```diff
@@ -140,6 +140,52 @@
     domain: java.lang
     type: GarbageCollector
     name: ZGC Pauses
+    attribute:
+      CollectionCount:
+        alias: jvm.gc.zgc_pauses_collection_count
+        metric_type: counter
```
vickenty, line 146: Since these are new metrics, would it make sense to use `monotonic_count` type here?

carlosroman: Sadly this isn't a new metric as reusing the old `jvm.gc.zgc_pauses_collection_count` that was created originally to support ZGC (added in #303).

PR #503, opened 2023-12-28
Adding Maven cache to CircleCI config

carlosroman:
Noticed some tests failing due to not being able to pull dependencies down. Adding a cache in hope that it will stop that issue coming up again.
Caching doesn't work for the misbehaving JMX server as that is built in Docker and currently no way to share dependencies with the build process in Container Tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,12 +2,37 @@ version: 2
 lint_steps: &lint_steps
   steps:
     - checkout
+    - restore_cache:
+        name: Restore Maven cache
+        keys:
+          # when pom.xml file changes, use increasingly general patterns to restore cache
+          # Linting has own cache as deps not downloaded during test phases. If doesn't exist will
+          # fall back on any available cache.
+          - &maven-cache maven-cache-{{ .Branch }}-{{ checksum "pom.xml" }}-lint
+          - maven-cache-{{ .Branch }}-
+          - maven-cache-
     - run: ./mvnw verify -B -Dhttps.protocols=TLSv1.2 -DskipTests
+    - save_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,12 +2,37 @@ version: 2
...
+        key: *maven-cache
+        paths:
+          - ~/.m2
 build_steps: &build_steps
   steps:
     - checkout
+    - restore_cache:
+        name: Restore Maven cache
+        keys:
+          # when pom.xml file changes, use increasingly general patterns to restore cache
```
scottopell, line 25: could you explain this a little? The save-cache step uses the `maven-cache` anchor, which appears to be defined here.

So the value for `maven-cache` is (for example) `maven-cache-master-ab2c3ffef33`. What are the other 2 below this then?

scottopell (approved):

.circleci/config.yml
carlosroman: The other two are there incase this is a brand new branch. If the branch is new it will then try and find the latest cache that begins `maven-cache-{{ .Branch }}-` and if that fails then just grabs the latest version of the cache that maches `maven-cache-` (i.e. any branch). It then will save the cache after the first run as `maven-cache-{{ .Branch }}-{{ checksum "pom.xml" }}-lint` as that is what the `maven-cache` anchor value is.

scottopell: ooooh, thanks, that is helpful. So the anchor value for `maven-cache` is always the first (most specific) key. The cache-loading is circle-ci functionality that will pick the best key, but when it goes to save the key, it will save with the most specific key.

PR #505, opened 2023-12-28
AMLII-1360: Creates telemetry bean for Application-level metrics

scottopell:
This utilizes the existing `jmxfetch_telemetry_instance` that was configured as part of #467  and establishes a similar bean for Application level metrics of interest.

Initially this will be 3 metrics:
- count of currently running instances
- count of currently broken instances
- count of broken instance events that have occurred.


rayz (approved):
code LGTM. Will `agent status` be updated to display these stats?

scottopell:
Great question, thanks for bringing that up @rayz 

I'll create a card to add these to the Status page as well

PR #506, opened 2023-12-28
Adds maven-compiler-plugin and turns on linter warnings

scottopell:
Currently have mostly unchecked-cast warnings, but a few others mixed in. This change should not go in without fixing the current linter errors _and_ enforcing this via CI


scottopell:
Closing for now, can be taken up again in the future.

PR #514, opened 2024-01-19
AMLII-1420: Implements more robust container wait strategy

scottopell:
This implements a more complex "wait strategy" for the test-containers framework to handle all environments where we need to execute these tests.
In particular, the gitlab environment where we execute these will never have a passing port check, the networking used there does not support the "is listening" check provided by the testcontainers framework.

So the desired behavior is a combination of log message watching and "is listening" port check.
The idea here is that the log message can sometimes signal _too early_. The reason for this is not quite clear, but given that we fail with "No route to host" when the log message signals, I suspect that the networking stack has not finished initializing.
The port check, if it succeeds, will always signal correctly, so the final strategy looks like this:
```
waitFor port_listening OR (log_msg AND 5-second timer)
```

The log has an added timer that will act as an artificial delay to account for the fact that log messages can signal too early.


rayz (approved):
lgtm 👍

carlosroman (approved):

PR #522, opened 2024-05-30
[AMLII-1815] Telemetry is no longer initialized when telemetry disabled

carlosroman:
Even when telemetry is disabled the JMXFetch will still register an MBean server. This causes issues when run in embedded mode where there might be a custom MBean manager. This was picked up in tests [here](https://github.com/DataDog/dd-trace-java/pull/6935#issuecomment-2139069627).


maycmlee (approved):

rayz (approved):
looks good to me

PR #529, opened 2024-06-10
[AMLII-1828] Cleaning up pom.xml

carlosroman:
Cleaned up the `pom.xml` and bumped some of the dependencies. Deployments still work after all the changes (see `SNAPSHOT` [here](https://oss.sonatype.org/content/repositories/snapshots/com/datadoghq/jmxfetch/0.49.3-SNAPSHOT/jmxfetch-0.49.3-20240710.153833-10.pom)).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
         <javax.management.j2ee-api.version>1.1.2</javax.management.j2ee-api.version>
         <jcommander.version>1.35</jcommander.version>
         <junit.version>4.13.2</junit.version>
-        <lombok.version>1.18.22</lombok.version>
+        <lombok.version>1.18.32</lombok.version>
         <mockito.version>2.28.2</mockito.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -95,7 +100,7 @@
                     <plugin>
                         <groupId>org.sonatype.plugins</groupId>
                         <artifactId>nexus-staging-maven-plugin</artifactId>
-                        <version>1.6.13</version>
+                        <version>${nexus-staging-maven-plugin.version}</version>
```
vickenty, line 103: Just curious, what is the advantage of specifying the version via a property?

vickenty (approved):

pom.xml
carlosroman: I went back and forth between specifying them in property or just moving them out of property completely or moving it all to `dependency-management` section. I was aiming for some consistency while I went through and upgraded some of the plugins and dependencies as we had mixture.

PR #537, opened 2024-08-23
`max_returned_metrics`: parse string value with fallback

tbavelier:
### What does this PR do ?
This PR allows the use of integer string value for the `max_returned_metrics` parameter. If the string cannot be parsed, falls back to default limit instead.

### Motivation
Recent customer case where their JMX configuration would not work once they added `max_returned_metrics`. As they were using k8s annotations, they tried to use a string in the JSON instead of the integer, causing the instance not to be loaded with the error message `Unable to create instance. Please check your yaml file` : 
```json
{
  "instances": [
    {
      "host": "%%host%%",
      "port": 9999,
      "max_returned_metrics": "2000"
    }
  ]
}
```
instead of
```json
{
  "instances": [
    {
      "host": "%%host%%",
      "port": 9999,
      "max_returned_metrics": 2000
    }
  ]
}
```
With this PR, both syntaxes would work.

### DIFF
--- a/src/main/java/org/datadog/jmxfetch/Instance.java
+++ b/src/main/java/org/datadog/jmxfetch/Instance.java
@@ -164,7 +164,9 @@ public class Instance {
             try {
                 this.maxReturnedMetrics = Integer.parseInt((String) maxReturnedMetrics);
             } catch (NumberFormatException e) {
-                log.warn("Cannot convert max_returned_metrics to integer in your instance configuration. Defaulting to '{}'.", MAX_RETURNED_METRICS);
+                log.warn(
+                    "Cannot convert max_returned_metrics to integer in your instance configuration. Defaulting to '{}'.",
+                    MAX_RETURNED_METRICS);
                 this.maxReturnedMetrics = MAX_RETURNED_METRICS;
             }
         } else {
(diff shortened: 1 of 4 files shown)


src/test/java/org/datadog/jmxfetch/TestInstance.java
```diff
@@ -97,6 +97,31 @@ public void testBaselineDefaultHostname() throws Exception {
         }
     }
 
+    @Test
```
carlosroman, line 100: Thank you ❤️

src/main/java/org/datadog/jmxfetch/Instance.java
```diff
@@ -160,6 +160,13 @@ public Instance(
         Object maxReturnedMetrics = this.instanceMap.get("max_returned_metrics");
         if (maxReturnedMetrics == null) {
             this.maxReturnedMetrics = MAX_RETURNED_METRICS;
+        } else if (maxReturnedMetrics instanceof String) {
+            try {
+                this.maxReturnedMetrics = Integer.parseInt((String) maxReturnedMetrics);
+            } catch (NumberFormatException e) {
+                log.warn("Cannot convert max_returned_metrics to integer in your instance configuration. Defaulting to '{}'.", MAX_RETURNED_METRICS);
```
carlosroman, line 167: This line might break the linter (waiting for it to run) so you may want to do something like this (I probably have the spaces all wrong):
```suggestion
                log.warn(
                    "Cannot convert max_returned_metrics to integer in your instance configuration. Defaulting to '{}'.",
                     MAX_RETURNED_METRICS);
```

carlosroman (approved):

tbavelier:
Superseded by https://github.com/DataDog/jmxfetch/pull/539

PR #608, opened 2026-04-10
feat(ai): Adding AGENTS.md/CLAUDE.md and new skill to create GC tests (including adding missing GC tests).

carlosroman:
Adding some coding agent files.

Used the new skill `add-gc-test` to add missing GC tests.


vickenty (approved):
I wouldn't expect code changes from a PR with this title, even if it is just the tests. Would it make sense to update PR to highlight actual code changes as well?

carlosroman:
@vickenty Good catch. Will update the title to reflect I added some new tests.

